PR #1, opened 2022-08-23
Fix some issues related to circular import

MaferMazu:
# Description

This PR fix the circular import error; and LTIProviderConfig and _apply_saml_overrides imports.

I made two changes:
- Implement get_registration_extension_form instead of importing it
- Check if the custom form exists before using it

## Related
- https://plataforma-nau.atlassian.net/browse/FAN-49


igobranco (approved):

felipemontoya (approved):
I think we can merge this into nau/lilac.master and continue building from there

PR #4, opened 2022-08-29
fix: add openedx-filters and constrain that pip-tools<5.4 and click==…

MaferMazu:
## Description

This PR adds `openedx-filters` as a plugin requirement, and solves the problem that when doing `make upgrade` pip-tools was in advanced version and click too.

## Solution
Restricted pip-tools<5.4 and click==7.1.2 as is on the platform.

### Reference
- [Pip-tools<5.4 contrain on the platform](https://github.com/fccn/edx-platform/blob/nau/lilac.master/requirements/constraints.txt#L80)
- [Click==7.1.2 on the platform](https://github.com/fccn/edx-platform/blob/nau/lilac.master/requirements/edx/base.txt#L41)


felipemontoya:
This looks good. Did you run the make upgrade from a container running bash? Or outside of the container?

felipemontoya (approved):

MaferMazu:
Inside the container @felipemontoya

PR #5, opened 2022-08-30
feat: add filters to restrict enrollment FAN-41

MaferMazu:
## Description

This PR adds the feature [FAN-41: NAU Academies - restrict course enrollment by email domain](https://plataforma-nau.atlassian.net/jira/software/projects/FAN/boards/26?selectedIssue=FAN-41).

Screenshot from 2022-09-26 13-30-46
_User interface when a user can't enroll (en)_
Screenshot from 2022-09-26 13-31-05
_User interface when a user can't enroll (pt-pt)_

Screenshot from 2022-08-29 16-53-49
_Internal logs_

## Dependencies
- fccn/edx-platform [PR #1](https://github.com/fccn/edx-platform/pull/1)
- fccn/edx-platform [PR #4](https://github.com/fccn/edx-platform/pull/4) 

## How to test

In #6 I added tests.

In tutor: 
- Use a tutor environment with edx-platform with `nau/lilac.master` branch
- Use nau-openedx-extension (this branch)
- You need to add the following settings
```
FEATURES.update({
    "ENABLE_OTHER_COURSE_SETTINGS": True
    })
OPEN_EDX_FILTERS_CONFIG = {
        "org.openedx.learning.course.enrollment.started.v1": {
            "fail_silently": False,
            "pipeline": [
                "nau_openedx_extensions.filters.pipeline.FilterEnrollmentByDomain",
            ],
        },
    }
```
- Create a course on the platform
- Go to studio > the course > settings > advanced settings (example: http://studio.lilac.edunext.link:8001/settings/advanced/course-v1:edX+DemoX+Demo_Course) and add in other course settings:
```
{
    "filter_enrollment_by_domain_list": [
        "domain1.com",
        "domain2.com",
        "example.com"
    ]
}
```
- Create a user with a domain or subdomain in the filter_enrollment_by_domain_list, and the user will be able to enroll
- Create a user with a domain, not in the list, and the user won't be able to enroll
- You can change the lang preference, and the UI message error will be in the correct lang


nau_openedx_extensions/filters/models.py
```diff
@@ -0,0 +1,23 @@
+"""Models used in openedx filters."""
+
+from django.db import models
+from opaque_keys.edx.django.models import CourseKeyField
+
+
+class DomainsAllowedPerCourse(models.Model):
```
felipemontoya, line 7: I am more in favour of putting this into the [advanced course settings](https://github.com/openedx/edx-platform/blob/d9f54b9f460bed0001e999dac8884d3b8e8a949f/xmodule/course_module.py#L1021) in studio. It prevents the need for a custom model in mysql. You can access this via studio and therefore you can make the course authors use it.

@igobranco @sandroscosta  what would be your preference here? I think we have used "other_course_settings" in the past with Nau

nau_openedx_extensions/filters/pipeline.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+
+    def run_filter(self, user, course_key, mode):   # pylint: disable=unused-argument, no-self-use
+        """Filter."""
+
+        try:
+            course_and_domains = DomainsAllowedPerCourse.objects.get(course_key=course_key)
+            user_domain = user.email.split("@")[1]
+            if user_domain not in course_and_domains.get_allowed_domains_list:
+                raise CourseEnrollmentStarted.PreventEnrollment("You can't enroll on this course.")
```
felipemontoya, line 34: The message could be more specific.

"You can't enroll on this course because your email domain is not allowed. If you think this is an error, contact the course support"

nau_openedx_extensions/filters/models.py
igobranco: @felipemontoya  and @MaferMazu  I didn't know the existence of the "other_course_settings" field.
We don't have it activated on NAU.
I have the same opinion @felipemontoya I think it is better to use "other_course_settings" on the Course Studio advanced settings.

nau_openedx_extensions/filters/pipeline.py
igobranco: We also need that message to be translated.

MaferMazu: Here I am going to note down the use of wildcards: https://docs.python.org/3/library/fnmatch.html

And regarding having a more descriptive message, that message is printed on the server and not on the user side, I have to see how to put it on the user side, but I don't think it will be a problem.

felipemontoya: The UI shows the message from the server, which is relied by the http request. We do need to make the message descriptive. Translation is not an issue, we put it in the catalog and done.

nau_openedx_extensions/filters/pipeline.py
```diff
@@ -0,0 +1,35 @@
...
+        }
+    """
+
+    def run_filter(self, user, course_key, mode):   # pylint: disable=unused-argument, no-self-use
+        """Filter."""
+
+        user_domain = user.email.split("@")[1]
+        other_course_settings = get_other_course_settings(course_key)
+        domains_allowed = other_course_settings.get("domains_allowed", [])
+        if domains_allowed and user_domain not in domains_allowed:
```
felipemontoya, line 33: Will you include the fnmatch comparison here?

https://docs.python.org/3/library/fnmatch.html

nau_openedx_extensions/filters/pipeline.py
```diff
@@ -0,0 +1,35 @@
...
+            }
+        }
+    """
+
+    def run_filter(self, user, course_key, mode):   # pylint: disable=unused-argument, no-self-use
+        """Filter."""
+
+        user_domain = user.email.split("@")[1]
+        other_course_settings = get_other_course_settings(course_key)
+        domains_allowed = other_course_settings.get("domains_allowed", [])
```
felipemontoya, line 32: In the context of the code "domain_allowed" looks like it is enough information, but at the course config it might not be.

I would call it "filter_enrollment_by_domain_list".
So you know it is a list what this is expecting and you also have the name of the filter.

MaferMazu:
This PR is ready for review.
- The fnmatch for subdomains was added.
- The exception message can be showed in the user interface with [this PR #4 in edx-platform](https://github.com/fccn/edx-platform/pull/4)
- And I add unit test for this functionality and for the wrapper in this PR #6 

If you are okey I am going to squash the commits in this PR and merge when I have the approval

@felipemontoya @igobranco @MoisesGSalas

felipemontoya (approved):
I think this covers my previous comments. Thanks @MaferMazu

PR #6, opened 2022-09-15
test: add wrapper and filter tests FAN-58

MaferMazu:
## Description

This PR adds tests for the PR #5 

- Add pytest and coverage to test.in and update requirements (I use python 3.8, pip 20.0.1, and pip-tools 5.3.1).
- Add support for python 3.10 in tox.
- Add backends for tests.
- Add wrappers test.
- Add pipeline module unit tests.
- Correct the test execution by creating an email_module_l_v1 backend and adding it to the plugin settings.
- Add some pylint disable comments.

## How to test

- In the terminal run
`export DJANGO_SETTINGS_MODULE='nau_openedx_extensions.settings.test'`
- Then, make tests
`make python-test`

## Notes for next steps and later improvements

- Run quality test returns an error because there are a lot of Pylint errors (If you want to test quality, you can use `pylint <file-name>`)
- To avoid exporting DJANGO_SETTINGS_MODULE when running unit tests and set it automatically, you need to do it in the [Pytest configurations files](https://docs.pytest.org/en/7.1.x/reference/customize.html#configuration). I can't use [pytest-django](https://pytest-django.readthedocs.io/en/latest/) because some models import Openedx platform features directly and don't have a backend in the wrappers. 
- Drop support for Django 22


felipemontoya:
@MaferMazu do you run this tests from the tutor dev bash? Or are you running the out of the containers?

MaferMazu:
@felipemontoya I ran it out of the container

felipemontoya (approved):
I think this is better to merge this sooner rather than waiting for long with the branch open

PR #17, opened 2023-03-13
fix: filter enrollment by domain

igobranco:
Fix subdomain matcher.
GN-1129

### DIFF
--- a/nau_openedx_extensions/filters/pipeline.py
+++ b/nau_openedx_extensions/filters/pipeline.py
@@ -30,12 +30,12 @@ class FilterEnrollmentByDomain(PipelineStep):   # pylint: disable=too-few-public
     def run_filter(self, user, course_key, mode):   # pylint: disable=unused-argument, arguments-differ
         """Filter."""
 
-        user_domain = user.email.split("@")[1]
+        user_domain = user.email.split("@")[1].lower()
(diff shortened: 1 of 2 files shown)


nau_openedx_extensions/filters/pipeline.py
```diff
@@ -30,12 +30,12 @@ class FilterEnrollmentByDomain(PipelineStep):   # pylint: disable=too-few-public
...
         """Filter."""
 
-        user_domain = user.email.split("@")[1]
+        user_domain = user.email.split("@")[1].lower()
         other_course_settings = get_other_course_settings(course_key)
         domains_allowed = other_course_settings.get("value", {}).get("filter_enrollment_by_domain_list", [])
         allowed = False
         for domain in domains_allowed:
-            if fnmatch(user_domain, f"*{domain}"):
+            if user_domain == domain or fnmatch(user_domain, f"*.{domain}"):
```
sandroscosta, line 38: doesn't `fnmatch` do the same as `user_domain == domain`?

igobranco: No. One is the exactly match the other is for subdomain.
The old code this situation would match
If the allowed domain had:
- ina.pt
would match `anotherdomainina.pt` and that is incorrect.

I've added a new test to check this situation.

PR #33, opened 2023-08-29
Ensure SAML users have a usable password

sandroscosta:
FAN-155

This pipeline function assigns an usable password to an user in case that the user has an unusable password on user creation.
At the creation of new users through some TPA providers, some of them are created with an unusable password, a user with an unusable password cannot login properly in the platform if the common.djangoapps.third_party.pipeline.set_logged_in_cookies step is enabled.

The command should run after `social_core.pipeline.user.create_user` on the `SOCIAL_AUTH_TPA_SAML_PIPELINE`.

This should solve the issues with the SAML authentication

@igobranco: Please review this PR. There might be some issues with some imports or dependencies that I was no able to solve. We should also test this in STAGE. Google accounts can be enabled for this purpose.

### DIFF
--- a/nau_openedx_extensions/third_party_auth/pipeline.py
+++ b/nau_openedx_extensions/third_party_auth/pipeline.py
@@ -3,9 +3,21 @@ Auth pipeline definitions.
 """
 from __future__ import absolute_import, unicode_literals
 
+import logging
+
 from django.conf import settings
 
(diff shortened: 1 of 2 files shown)


igobranco:
Closing, replaced in favor of https://github.com/fccn/edx-platform/pull/13

PR #57, opened 2024-12-12
feat: add `allow_newsletter` field to account settings

BryanttV:
### Description
This PR adds the `allow_newsletter` field _(Checkbox)_ in the Account Settings. This allows each user to update the value of the field.

### Related Issues
- https://github.com/fccn/nau-technical/issues/279
- https://github.com/eduNEXT/consulting-issues-mapping/issues/10

### Testing Instructions
1. Create a Tutor environment in the Redwood version.
2. Install this plugin with the changes in this PR.
3. Create a mount of `edx-platform` with the changes in [this PR](https://github.com/fccn/edx-platform/pull/19).
4. Update the `account.redirect_to_microfrontend` Waffle Flag. Set `Everyone` to `No`. This allows access to the legacy interface.
5. Go to `{lms_domain}/account/settings`
6. You should see the new checkbox field in the **Additional Information** section.
7. Try to update the field and verify that the value was updated in the model.

### Screenshots
image

### DIFF
--- a/nau_openedx_extensions/custom_registration_form/context_extender.py
+++ b/nau_openedx_extensions/custom_registration_form/context_extender.py
@@ -27,7 +27,7 @@ def get_fields(custom_model_instance):
     for field in custom_fields:
         if field.name not in allowed_fields:
             continue
-        if isinstance(field, (models.CharField, models.TextField)):
+        if isinstance(field, (models.CharField, models.TextField, models.BooleanField)):
             yield field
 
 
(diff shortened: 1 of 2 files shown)


MaferMazu (approved):
Thanks for this, @BryanttV, this worked as expected ✨

My only suggestion, which is not a blocker, is that although this feature is related to the scheduled nudged emails, it does not have a direct dependency. Therefore, to ensure clarity and that the PRs do not block each other, I recommend that the PR go to `redwood.master` instead of the nudged branch.

igobranco (approved):
LGTM ✔️

PR #78, opened 2025-07-03
feat: add course certificate dispatch system with configurable external services

efortish:
Add course certificate dispatch system with configurable external services
=========================================================================

Summary
-------
This PR introduces a new Django management command system for dispatching course certificates to external services based on YAML configuration. The system provides a flexible, configurable way to send certificate data to multiple external APIs with custom field mapping and transformations.

Key Features
------------
- **Configurable external services**: Define multiple services in YAML with different endpoints, authentication, and field mappings
- **Field extraction system**: Modular extractors for student, course, certificate data, etc
- **Data transformations**: Support for MD5 hashing, base64 encoding, and other transformations
- **Flexible authentication**: Support for Bearer, Basic, API Key, and custom authentication types
- **Batch processing**: Efficient pagination and batch processing of certificates
- **Dry-run mode**: Test configuration and see generated payloads without sending actual requests
- **Service filtering**: Process all services or target specific ones
- **Async support**: Foundation for Celery-based asynchronous processing (commented for future implementation)

Components Added
----------------

1. Django Management Command

- ``nau_openedx_extensions/coursecertificate/management/commands/send_certificates_by_web_service.py``
- Processes certificates from the last N days
- Supports both synchronous and asynchronous modes
- Includes comprehensive error handling and logging

2. Engine
- ``engine.py`` has the logic to process the dispatch system.
- The normal workflow would be something like ``command`` ---> ``engine.py`` ---> ``output``
- When async flag, the worflow would be   ``command`` ---> ``tasks.py`` ---> ``engine.py`` ---> ``output`` (not implemented yet)

---
3. Field Extractors

- ``nau_openedx_extensions/coursecertificate/extractors.py`` - certificate_date, certificate_url, course_id, course_code, course_name, student_email. student_username, student_name, student_nau_user_extended_model_field, student_enrolled_date

---
4. Configuration System

- ``nau_openedx_extensions/coursecertificate/config.yml`` - YAML configuration for external services
- Support for multiple services with different configurations
- Configurable field mappings and transformations per service

---
5. Django App Configuration

- ``nau_openedx_extensions/coursecertificate/apps.py`` - Django app configuration with plugin support
- Proper integration with Open edX plugin architecture

---
Usage Examples
```
   # Process all certificates services
   python manage.py lms send_certificates_by_web_service

   # Dry run mode to see what would be sent
   python manage.py lms send_certificates_by_web_service --dry-run

   # Process specific service only
   python manage.py lms send_certificates_by_web_service --service-name portugal_digital

   # Custom time range and batch size
   python manage.py lms send_certificates_by_web_service --days 30 --page-size 50
```
Configuration Format example:
--------------------

```
   NAU_SEND_COURSE_CERTIFICATE_CONFIG:
  - service_name: portugal_digital
    endpoint_url: https://...
    auth_token: XXXXXXXXXXXXXXXXXXXXX
    auth_type: bearer
    auth_header: Authorization
    page_size: 100
    days: 30
    fields:
      - name: key
        func: nau_openedx_extensions.coursecertificate.extractors.student.email
        trans: md5
      - name: value
        func: nau_openedx_extensions.coursecertificate.extractors.course.id
```

Future Enhancements
-------------------
*(prepared but not implemented)*

- Celery-based asynchronous processing per service
- Certificate-specific listeners for real-time processing
- Payload caching system

### Related Issues
https://github.com/fccn/nau-technical/issues/555

### DIFF
--- a/nau_openedx_extensions/coursecertificate/engine.py
+++ b/nau_openedx_extensions/coursecertificate/engine.py
@@ -186,11 +186,11 @@ class CertificateEngine:
 
                 if args:
                     if isinstance(args, list):
-                        filtered_certificates = filter_func(filtered_certificates, *args)
+                        filtered_certificates = filter_func(filtered_certificates, *args)  # pragma: no cover
                     else:
                         filtered_certificates = filter_func(filtered_certificates, args)
                 else:
-                    filtered_certificates = filter_func(filtered_certificates)
+                    filtered_certificates = filter_func(filtered_certificates)  # pragma: no cover
 
             except (ImportError, AttributeError, TypeError) as exc:
                 self.log(f"Error applying filter {filter_config['func']}: {exc}")
@@ -404,19 +404,29 @@ class CertificateEngine:
 
         return success
 
-    def get_certificates_queryset(self, days: int) -> QuerySet:
+    def get_certificates_queryset(self, days: int, certificate_id: int | None) -> QuerySet:
         """
-        Get certificates queryset filtered by date.
-
-        Retrieves certificates created within the specified number of days from now.
-        Uses read replica if available for better performance.
+        Get certificates queryset filtered by date and optionally by certificate ID.
 
         Args:
             days (int): Number of days to look back from current date.
+            certificate_id (int | None): Specific certificate ID to filter by.
+                If provided, only this certificate will be returned regardless of date.
 
         Returns:
             QuerySet: Filtered and ordered certificates queryset with user relations.
         """
+        if certificate_id:
+            queryset = use_read_replica_if_available(
+                GeneratedCertificate.objects.filter(id=certificate_id).select_related("user")
+            )
+
+            if not queryset.exists():
+                self.log(f"WARNING: Certificate with ID {certificate_id} does not exist.")
+                return GeneratedCertificate.objects.none()
+
+            return queryset
+
         begin_date = datetime.now(UTC) - timedelta(days=days)
         return use_read_replica_if_available(
             GeneratedCertificate.objects.filter(created_date__gte=begin_date)
@@ -429,7 +439,7 @@ class CertificateEngine:
         Extract and validate processing parameters.
 
         Combines command-line options with service configuration defaults to determine
-        processing parameters like days, page size, and dry run mode.
+        processing parameters like days, page size, certificate ID, and dry run mode.
 
         Args:
             service_config (dict): Service configuration containing:
@@ -438,12 +448,14 @@ class CertificateEngine:
             options (dict): Command options containing:
                 - days (int, optional): Override for days parameter
                 - page_size (int, optional): Override for page size
+                - certificate_id (int, optional): Specific certificate ID to process
                 - dry_run (bool, optional): Dry run mode flag
 
         Returns:
             dict: Dictionary containing validated processing parameters:
                 - days (int): Number of days to process
                 - page_size (int): Page size for pagination
+                - certificate_id (int, optional): Specific certificate ID
                 - dry_run (bool): Dry run mode flag
         """
         days = options.get("days")
@@ -454,7 +466,12 @@ class CertificateEngine:
         if page_size is None:
             page_size = service_config.get("page_size", 1000)
 
-        return {"days": days, "page_size": page_size, "dry_run": options.get("dry_run", False)}
+        return {
+            "days": days,
+            "page_size": page_size,
+            "certificate_id": options.get("certificate_id"),
+            "dry_run": options.get("dry_run", False),
+        }
 
     def _process_certificates_page(self, certificates, service_config: dict, dry_run: bool) -> bool:
         """
@@ -537,6 +554,7 @@ class CertificateEngine:
             options (dict): Processing options containing:
                 - days (int, optional): Override for days parameter
                 - page_size (int, optional): Override for page size
+                - certificate_id (int, optional): Specific certificate ID to process
                 - dry_run (bool, optional): Dry run mode flag
         """
         service_name = service_config.get("service_name")
@@ -544,10 +562,13 @@ class CertificateEngine:
 
         params = self._get_processing_parameters(service_config, options)
 
-        self.log(f"Processing certificates from the last {params['days']} days")
-        self.log(f"Page size: {params['page_size']}")
+        if params["certificate_id"]:
+            self.log(f"Processing specific certificate with ID: {params['certificate_id']}")
+        else:
+            self.log(f"Processing certificates from the last {params['days']} days")
+            self.log(f"Page size: {params['page_size']}")
 
-        certificates_queryset = self.get_certificates_queryset(params["days"])
+        certificates_queryset = self.get_certificates_queryset(params["days"], params["certificate_id"])
         certificates_queryset = self.apply_filters(certificates_queryset, service_config)
 
         self._process_certificates_pages(certificates_queryset, service_config, params)
--- a/nau_openedx_extensions/coursecertificate/engine.py
+++ b/nau_openedx_extensions/coursecertificate/engine.py
@@ -1,0 +1,553 @@
+"""
+Certificate processing engine.
+
+This module provides a comprehensive certificate processing engine that handles
+the extraction, transformation, and delivery of course certificates to external
+services. It includes functionality for filtering certificates, applying data
+transformations, and sending them to configured endpoints with various
+authentication methods.
+
+The engine is designed to be independent of Django commands and Celery tasks,
+making it easily testable and reusable across different contexts.
+
+Classes:
+    CertificateEngine: Main engine class that handles all certificate processing logic.
+"""
+
+import base64
+import hashlib
+import importlib
+import json
+from datetime import datetime, timedelta
+from typing import Callable, Dict, List, Optional
+
+import requests
+from django.core.paginator import Paginator
+from django.db.models import QuerySet
+from pytz import UTC
+
+from nau_openedx_extensions.edxapp_wrapper.certificates import GeneratedCertificate
+from nau_openedx_extensions.edxapp_wrapper.util import use_read_replica_if_available
+
+
+class CertificateEngine:
+    """
+    Certificate processing engine - handles all business logic.
+
+    This engine provides a complete solution for processing course certificates
+    and sending them to external services. It supports multiple authentication
+    methods, data transformations, filtering, and paginated processing.
+
+    The engine is designed to be independent of Django commands and Celery tasks,
+    making it easily testable and reusable across different contexts.
+
+    Attributes:
+        VALID_AUTH_TYPES (List[str]): Supported authentication types.
+        log (Callable[[str], None]): Logger function for output messages.
+    """
+
+    VALID_AUTH_TYPES = ["bearer", "basic", "api_key"]
+
+    def __init__(self, logger: Optional[Callable[[str], None]] = None):
+        """
+        Initialize the engine with an optional logger function.
+
+        Args:
+            logger (Optional[Callable[[str], None]]): Function to log messages.
+                If not provided, defaults to print function.
+        """
+        self.log = logger or print
+
+    def _transform_md5(self, value: str) -> str:
+        """
+        Transform value using MD5 hash.
+
+        Args:
+            value (str): The value to transform.
+
+        Returns:
+            str: MD5 hash of the input value as hexadecimal string.
+        """
+        return hashlib.md5(str(value).encode()).hexdigest()
+
+    def _transform_base64(self, value: str) -> str:
+        """
+        Transform value using Base64 encoding.
+
+        Args:
+            value (str): The value to transform.
+
+        Returns:
+            str: Base64 encoded string of the input value.
+        """
+        return base64.b64encode(str(value).encode()).decode()
+
+    def apply_transformations(self, value: str, trans: str) -> str:
+        """
+        Apply transformations to a field value.
+
+        Supports MD5 hash and Base64 encoding transformations.
+        The transformation type is case-insensitive and leading whitespace is stripped.
+
+        Args:
+            value (str): The value to transform.
+            trans (str): The transformation type ('MD5' or 'BASE64').
+
+        Returns:
+            str: The transformed value.
+
+        Raises:
+            ValueError: If the transformation type is not supported.
+        """
+        transformations = {
+            "MD5": self._transform_md5,
+            "BASE64": self._transform_base64,
+        }
+
+        transform_func = transformations.get(trans.strip().upper())
+        if transform_func:
+            return transform_func(value)
+        raise ValueError(f"Encryption method not valid: {trans}")
+
+    def extract_field_value(self, certificate: QuerySet, field_config: dict) -> Optional[str]:
+        """
+        Extract field value from certificate using configured function.
+
+        Dynamically imports and calls the extraction function specified in the field
+        configuration. Supports passing arguments to the function and applying
+        transformations to the result.
+
+        Args:
+            certificate (QuerySet): The certificate object to extract data from.
+            field_config (dict): Configuration dictionary containing:
+                - func (str): Module path and function name (e.g., 'module.function')
+                - args (list, optional): Arguments to pass to the function
+                - trans (str, optional): Transformation to apply to the result
+                - name (str): Field name for error reporting
+
+        Returns:
+            Optional[str]: The extracted and optionally transformed field value,
+                or None if extraction fails.
+        """
+        func_path = field_config["func"]
+        args = field_config.get("args", [])
+        trans = field_config.get("trans")
+
+        try:
+            module_path, func_name = func_path.rsplit(".", 1)
+            module = importlib.import_module(module_path)
+            extract_func = getattr(module, func_name)
+
+            if args:
+                if isinstance(args, list):
+                    value = extract_func(certificate, *args)
+                else:
+                    value = extract_func(certificate, args)
+            else:
+                value = extract_func(certificate)
+
+            if trans:
+                value = self.apply_transformations(value, trans)
+
+            return value
+        except (ImportError, AttributeError, TypeError) as exc:
+            self.log(f"Error extracting field {field_config['name']}: {exc}")
+            return None
+
+    def apply_filters(self, certificates: QuerySet, service_config: dict) -> QuerySet:
+        """
+        Apply filters to certificates based on service configuration.
+
+        Dynamically imports and applies filter functions specified in the service
+        configuration. Each filter function receives the certificates queryset and
+        optional arguments, and returns a filtered queryset.
+
+        Args:
+            certificates (QuerySet): The certificates queryset to filter.
+            service_config (dict): Service configuration containing:
+                - filters (list, optional): List of filter configurations, each containing:
+                    - func (str): Module path and function name
+                    - args (list, optional): Arguments to pass to the filter function
+
+        Returns:
+            QuerySet: The filtered certificates queryset.
+        """
+        filters = service_config.get("filters", [])
+        filtered_certificates = certificates
+
+        for filter_config in filters:
+            try:
+                func_path = filter_config["func"]
+                args = filter_config.get("args", [])
+
+                module_path, func_name = func_path.rsplit(".", 1)
+                module = importlib.import_module(module_path)
+                filter_func = getattr(module, func_name)
+
+                if args:
+                    if isinstance(args, list):
+                        filtered_certificates = filter_func(filtered_certificates, *args)
+                    else:
+                        filtered_certificates = filter_func(filtered_certificates, args)
+                else:
+                    filtered_certificates = filter_func(filtered_certificates)
+
+            except (ImportError, AttributeError, TypeError) as exc:
+                self.log(f"Error applying filter {filter_config['func']}: {exc}")
+                continue
+
+        return filtered_certificates
+
+    def convert_certificates_to_service_format(self, certificates, service_config: dict) -> List[Dict]:
+        """
+        Convert certificates to the format required by the service.
+
+        Processes each certificate according to the field configuration specified
+        in the service config. Extracts configured fields and applies any
+        transformations before building the final data structure.
+
+        Args:
+            certificates: Iterable of certificate objects to convert.
+            service_config (dict): Service configuration containing:
+                - fields (list, optional): List of field configurations for extraction
+
+        Returns:
+            List[Dict]: List of dictionaries containing certificate data formatted
+                for the target service. Only certificates with successfully extracted
+                data are included.
+        """
+        fields_config = service_config.get("fields", [])
+        converted_data = []
+
+        for certificate in certificates:
+            cert_data = {}
+            for field_config in fields_config:
+                field_name = field_config["name"]
+                field_value = self.extract_field_value(certificate, field_config)
+                if field_value is not None:
+                    cert_data[field_name] = field_value
+
+            if cert_data:
+                converted_data.append(cert_data)
+
+        return converted_data
+
+    def _validate_service_config(self, service_config: dict) -> None:
+        """
+        Validate service configuration parameters.
+
+        Checks that the authentication type is supported and other required
+        configuration parameters are present.
+
+        Args:
+            service_config (dict): Service configuration containing:
+                - service_name (str, optional): Name of the service for error reporting
+                - auth_type (str, optional): Authentication type, defaults to 'bearer'
+
+        Raises:
+            ValueError: If auth_type is not in VALID_AUTH_TYPES.
+        """
+        service_name = service_config.get("service_name")
+        auth_type = service_config.get("auth_type", "bearer")
+
+        if auth_type not in self.VALID_AUTH_TYPES:
+            raise ValueError(
+                f"Invalid auth_type '{auth_type}' for service '{service_name}'. "
+                f"Valid options: {', '.join(self.VALID_AUTH_TYPES)}"
+            )
+
+    def _prepare_auth_headers(self, service_config: dict) -> dict:
+        """
+        Prepare authentication headers based on service configuration.
+
+        Builds HTTP headers with appropriate authentication based on the configured
+        authentication type. Supports Bearer tokens, Basic authentication, and API keys.
+
+        Args:
+            service_config (dict): Service configuration containing:
+                - auth_token (str, optional): Authentication token
+                - auth_type (str, optional): Authentication type, defaults to 'bearer'
+                - auth_header (str, optional): Header name, defaults to 'Authorization'
+
+        Returns:
+            dict: Dictionary of HTTP headers including authentication if configured.
+        """
+        headers = {
+            "Accept": "application/json",
+            "Content-Type": "application/json",
+        }
+
+        auth_token = service_config.get("auth_token")
+        auth_type = service_config.get("auth_type", "bearer")
+        auth_header = service_config.get("auth_header", "Authorization")
+
+        if auth_token:
+            if auth_type == "bearer":
+                headers[auth_header] = f"Bearer {auth_token}"
+            elif auth_type == "basic":
+                credentials = base64.b64encode(auth_token.encode()).decode()
+                headers[auth_header] = f"Basic {credentials}"
+            elif auth_type == "api_key":
+                headers[auth_header] = auth_token
+
+        return headers
+
+    def _log_dry_run_info(self, service_config: dict, certificates_data: List[Dict]) -> None:
+        """
+        Log dry run information.
+
+        Outputs detailed information about what would be sent to the service
+        during a dry run, including endpoint URL, authentication details, timeout, and payload.
+
+        Args:
+            service_config (dict): Service configuration containing:
+                - endpoint_url (str): API endpoint URL
+                - endpoint_timeout (int, optional): Request timeout in seconds
+                - auth_token (str, optional): Authentication token
+                - auth_type (str, optional): Authentication type
+                - auth_header (str, optional): Authentication header name
+            certificates_data (List[Dict]): The certificate data that would be sent.
+        """
+        api_url = service_config["endpoint_url"]
+        timeout = service_config.get("endpoint_timeout", 60)
+        auth_token = service_config.get("auth_token")
+        auth_type = service_config.get("auth_type", "bearer")
+        auth_header = service_config.get("auth_header", "Authorization")
+
+        self.log(f"[DRY RUN] Would send to {api_url}")
+        self.log(f"[DRY RUN] Headers would include: {auth_type} authentication")
+        self.log(f"[DRY RUN] Auth header: {auth_header}")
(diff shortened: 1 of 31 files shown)


nau_openedx_extensions/coursecertificate/management/commands/send_certificates_by_web_service.py
```diff
@@ -0,0 +1,332 @@
...
+    Send course certificates to external services based on YAML configuration.
+    """
+
+    help = "Send course certificates to external services"
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.config = None
+        self.dry_run = False
+        self.async_mode = False
```
BryanttV: Why do we need the `init` method? The default values can be added directly to the argument.

nau_openedx_extensions/coursecertificate/management/commands/send_certificates_by_web_service.py
```diff
@@ -0,0 +1,332 @@
...
+        begin_date = datetime.now(UTC) - timedelta(days=days)
+        return use_read_replica_if_available(
+            GeneratedCertificate.objects.filter(created_date__gte=begin_date)
+            .order_by("created_date")
+            .select_related("user")
+        )
+
+    def process_service(self, service_config: dict, options: dict) -> None:
+        """Process certificates for a specific service"""
+        service_name = service_config.get("service_name") or service_config.get("service")
```
BryanttV: Why do we have two possible names?

nau_openedx_extensions/coursecertificate/services/engine.py
```diff
@@ -0,0 +1,3 @@
+"""
```
BryanttV: Can we remove this unused module?

BryanttV (commented):
Thanks @efortish! LGTM, I just have some small comments.

nau_openedx_extensions/coursecertificate/management/commands/send_certificates_by_web_service.py
efortish: The main reason is due initially the config.yml had the keyword 'service', unfortunately, that caused conflicts with Django core keywords, so I putted both options in development, I'll remove the 'service' options due 'service_name' will be used instead

nau_openedx_extensions/coursecertificate/services/engine.py
efortish: Sure

nau_openedx_extensions/coursecertificate/management/commands/send_certificates_by_web_service.py
```diff
@@ -0,0 +1,334 @@
...
+            raise CommandError(f"Error parsing YAML configuration: {exc}") from exc
+
+    def log_msg(self, msg: str) -> None:
+        """Log a message immediately"""
+        self.stdout.write(msg)
+        self.stdout.flush()
+
+    def apply_transformations(self, value: str, trans: str) -> str:
+        """Apply transformations to a field value"""
+        if trans == "md5":
```
Alec4r: We only have 2 possible transformations using md5 to base64 and we are already using "if and elif", maybe we could move this to a dictionary to make it easier to extend in the future

nau_openedx_extensions/coursecertificate/management/commands/send_certificates_by_web_service.py
```diff
@@ -0,0 +1,334 @@
...
+        self,
+        service_config: dict,
+        certificates_data: list[dict],
+        dry_run: bool = False
+    ) -> bool:
+        """Send certificates to external service"""
+        service_name = service_config["service_name"]
+        api_url = service_config["endpoint_url"]
+        auth_token = service_config.get("auth_token")
+        auth_type = service_config.get("auth_type", "bearer")
```
Alec4r: we should validate the value of auth_type, to avoid possible errors due to wrong spelling.

nau_openedx_extensions/coursecertificate/management/commands/send_certificates_by_web_service.py
```diff
@@ -0,0 +1,334 @@
...
+        auth_header = service_config.get("auth_header", "Authorization")
+
+        if dry_run:
+            self.log_msg(f"[DRY RUN] Would send to {api_url}")
+            self.log_msg(f"[DRY RUN] Headers would include: {auth_type} authentication")
+            self.log_msg(f"[DRY RUN] Auth header: {auth_header}")
```
Alec4r: Please double check if this is printing the auth_token, even during dry-run mode, for security reasons, we should avoid exposing sensitive data like tokens in logs even in non production runs maybe we should masking the token or omitting it entirely from the output

nau_openedx_extensions/coursecertificate/management/commands/send_certificates_by_web_service.py
efortish: Sound good. on it

nau_openedx_extensions/coursecertificate/management/commands/send_certificates_by_web_service.py
efortish: Sounds good, on it

nau_openedx_extensions/coursecertificate/management/commands/send_certificates_by_web_service.py
efortish: On it, I checked it and it is not printing the auth_token:
image

efortish: Besides that, I think someone could print the token if a log is added to show it, I think omitting it is the best solution, on it

nau_openedx_extensions/coursecertificate/management/commands/send_certificates_by_web_service.py
efortish: DOne !

nau_openedx_extensions/coursecertificate/management/commands/send_certificates_by_web_service.py
efortish: Done!

nau_openedx_extensions/coursecertificate/management/commands/send_certificates_by_web_service.py
efortish: Done!

Tiago-Salles:
## too many modules

Is it necessary to separate the fields of each object by modules? Can it be in the same file separated by classes?
## coursecertificate/filters.py

It is doing the filter by organization using the memory, with a for loop at runtime. Using Django ORM we can optimize the consult, and process only what matters at runtime. @igobranco is the filter by the queryset not that clean, at the point that it justifies to for loop a giant list multiple times in the workflow?

```
def certificate_by_org(certificates: QuerySet, org: str) -> list:

return [certificate for certificate in certificates if certificate.course_id.org == org]
```

## send_certificates_by_web_service

Why to not use Django ORM and leave it up to MySQL to handle the transformation from the original value to md5 or base64?
#### MD5
[MD5](https://docs.djangoproject.com/en/5.2/ref/models/database-functions/#md5) expression will return all the necessary fields of the query as md5 directly.

```
data= Model.objects.annotate(
    y=MD5('y')
)
```
#### base64
[Func](https://docs.djangoproject.com/en/5.2/ref/models/expressions/#func-expressions) expression will return all the  necessary fields of the query as base64 directly.

```
data = Model.objects.annotate(
    field=Func(
        F('field'),
        function='TO_BASE64',
        output_field=CharField()
    )
)
```

By doing it, we avoid more one process using for loop.

## increase the query 

```
def get_certificates_queryset(self, days: int) -> QuerySet:
    """Get certificates queryset filtered by date"""
    begin_date = datetime.now(UTC) - timedelta(days=days)
	
    return use_read_replica_if_available(
	GeneratedCertificate.objects.filter(created_date__gte=begin_date)
	.order_by("created_date")
	.select_related("user")
    )
```

Leaving it up to Django ORM to select the relevant registers is correct, not only because it avoids loading unnecessary data to the memory, but also because it is optimized to do this kind of tasks, consulting, selecting, and filtering values with efficiency is the role of the ORM.

## How about to implement something like this

Using this approach we query all the data with filters, applying encryption, and parsing the result to a [JSON parse](https://docs.djangoproject.com/en/5.2/ref/models/database-functions/#jsonobject) directly from the query. 

The result of this function should be the data the `send_certificates_to_service` expects, making it open to rename this method, could be something like `assembly_data_to_send`, `build_payload_to_send`, some name that makes clear what this method does.

```
from django.db.models import MD5, Func, CharField, F
from django.db.models.functions import JSONObject

# It implements `Func` to add the MySQL function `TO_BASE64`
class ToBase64(Func):
    function = 'TO_BASE64'
    output_field = CharField()


def get_certificates_queryset(self, filters, days: int) -> QuerySet:
    """Get complete certificates queryset"""

    # each service has the field
    field_configs = ...
    begin_date = datetime.now(UTC) - timedelta(days=days)

    # adds the date filter
    filters["created_date__gt"] = begin_date

    annotations = {}
    json_fields = {}
	
    # Add the fields to `annotations` dynamically with
    # MD5 or ToBase64 expressions to leave up to ORM
    # apply in all encryption for necessary registers
	
    for field_config in field_configs:
        field_name = field_config["name"]
	trans = field_config.get("trans")
	
	if trans:
	    if trans.upper() == "MD5":
	        annotations[field_name] = MD5(field_name)
	    elif trans.upper() == "BASE64":
		annotations[field_name] = ToBase64(field_name)
		
	# normalize to be serialized using `JSONObject`
	normalized_field_name = field_name.replace("__", "_")
	json_fields[normalized_field_name] = F(field_name)

    # Now the queryset will have a new field `data`
    # which will be a JSON parse already processed
    return use_read_replica_if_available(
        GeneratedCertificate.objects.filter(**filters)
	    .annotate(**annotations)
	    .annotate(data=JSONObject(**json_fields)) 
	    .order_by("created_date")
	    .select_related("user")
    )
```

By applying a similar solution, it would make the `config.yaml` decrease for this service, which is good, as small as possible. 

@BryanttV, @efortish is it possible? What are your visions about it?

efortish:
Hello @Tiago-Salles, thank you so much for your review. While you were doing the review, we made some changes — we've now separated the command logic into a new file named engine.py to support a more flexible architecture. We will take your feedback into consideration as we continue working on improvements.

nau_openedx_extensions/coursecertificate/engine.py
```diff
@@ -0,0 +1,257 @@
...
+            if args:
+                if isinstance(args, list):
+                    value = extract_func(certificate, *args)
+                else:
+                    value = extract_func(certificate, args)
+            else:
+                value = extract_func(certificate)
+
+            # Apply transformations if specified
+            if trans:
```
Tiago-Salles, line 149: Please, when it is possible, avoid this kind of approach. Calling a method inside another method when it is not exactly the same responsibility makes the code less readable. In this case we could create a method like

```
    def execute_fields_steps(self, certificate: QuerySet, field_config: dict):
        trans = field_config.get("trans")
        
        field_value = self.extract_field_value(certificate=certificate, field_config=field_config)
        
        if trans:
            field_value = self.apply_transformations(value=field_value, trans=trans)
        
        return field_value
```

and pass each operation step by step, as it was a cookbook. This approach can be explored from the first method, the one who starts the whole operation (`process_service`). By doing it we identify very easily all the operations that happen in this workflow.

Tiago-Salles:
> Hello @Tiago-Salles, thank you so much for your review. While you were doing the review, we made some changes — we've now separated the command logic into a new file named engine.py to support a more flexible architecture. We will take your feedback into consideration as we continue working on improvements.

Thank you @efortish, it indeed looks better. Which makes me to remember that I forgot mentioning in the first comment I did,  the importance of leaving only the command stuffs inside the command context. Django instructions, parsing parameters, executing the first method is the role of the command. Otherwise, validating the `config.yaml` and its information sounds be an engine responsibility.

BryanttV:
Hi @Tiago-Salles, once again, thank you for your feedback!

The approach you proposed seems interesting. We've been reviewing it to determine the best way to implement it, and we've some comments regarding it.

## Filters

Currently, two filters are proposed based on the requirements (`by_course_id_regex` and `by_org`). At first, we didn’t see a way to apply the `by_org` filter using Django ORM. However, after doing some testing, we managed to do it using a regex as well: https://github.com/fccn/nau-openedx-extensions/pull/78/commits/258b58ff5a91f48033376f0b9dbd1f32744f7e54, this allowed us to avoid an additional iteration.

## New Approach

The idea is to minimize in-memory iterations by delegating as much as possible to the Django ORM. However, we identified a few points we need to consider before fully committing to this approach.

### Computed Fields

Some of the fields required by the requirements are not directly available from the `GeneratedCertificate` model. Specifically:

* `course_name`: This value can be obtained via an additional query to the `CourseOverview` model. This could be done using a [`Subquery`](https://docs.djangoproject.com/en/5.2/ref/models/expressions/#subquery-expressions).
* `course_code` (also known as `course_number` or simply `course` in Open edX): This value is not directly available in the model, but we can extract it from the `course_id`. Maybe we can use a [`REGEXP_SUBSTR`](https://dev.mysql.com/doc/refman/8.4/en/regexp.html#function_regexp-substr) (?).
* `enrolled_date`: This value can be obtained via an additional query to the `CourseEnrollment` model. Again, possibly using a `Subquery`.
* `certificate_url`: This URL isn’t stored in the model. We only have access to the `verify_uuid` field, from which we can construct the URL using: `f"{settings.LMS_ROOT_URL}/certificates/{verify_uuid}"`. This could be implemented with a [`Concat`](https://docs.djangoproject.com/en/5.2/ref/models/database-functions/#concat).

To implement a solution that accounts for these fields, we’d need to modify the structure of `config.yml` so it becomes configurable when a field is computed, specifying from where and how to extract the value. We created a very _rough_ initial test with something like this:

```yaml
NAU_SEND_COURSE_CERTIFICATE_CONFIG:
  - service_name: portugal_digital
    endpoint_url: https://academiaportugaldigital.pt/api/api/Course/FinishedIntegrationNau
    auth_token: XXXXXXXXXXXXXXXXXXXXX
    auth_type: bearer
    auth_header: Authorization
    page_size: 100
    days: 30
    fields:
      - name: email
        model_field: user__email
        trans: md5
      - name: username
        model_field: user__username
      - name: name
        model_field: user__profile__name
      - name: nif
        model_field: user__nauuserextendedmodel__nif
      - name: certificate_date
        model_field: created_date
      - name: certificate_url
        model_field: verify_uuid
        trans: concat
        prefix: http://local.edly.io/certificates/
      - name: course_id
        model_field: course_id

      # TODO: This is not working yet
      # - name: course_name_field
      #   model_keys: [course_id]
      #   model_class: CourseOverview
      # - name: enrolled_date
      #   model_keys: [user, course_id]
      #   model_class: CourseEnrollment
      # - name: course_code_field
      #   model_field: course_id
      #   trans: regex_substring
      #   regex: (?<=course-v1:[^+]+\+)[^+]+
    filters:
      - func: nau_openedx_extensions.coursecertificate.filters.certificate_by_course_id_regex
        args: course-v1
      - func: nau_openedx_extensions.coursecertificate.filters.certificate_by_org
        args: EduNext

```

```python
def get_certificates_queryset(self, service_config: dict, days: int) -> QuerySet:
    """Get complete certificates queryset"""
    field_configs = service_config.get("fields", [])
    begin_date = datetime.now(UTC) - timedelta(days=days)

    filters = {"created_date__gt": begin_date}
    all_annotations = {}
    json_fields = {}

    for field_config in field_configs:
        field_name = field_config["name"]  # desired output name
        model_field = field_config["model_field"]  # model field path
        trans = field_config.get("trans")

        # Internal annotation alias to avoid conflicts
        annotation_alias = f"annotated__{field_name}"

        # Build the expression
        expr = F(model_field)
        if trans:
            trans = trans.upper()
            if trans == "MD5":
                expr = MD5(F(model_field))
            elif trans == "BASE64":
                expr = ToBase64(F(model_field))
            elif trans == "CONCAT":
                prefix = field_config.get("prefix", "")
                suffix = field_config.get("suffix", "")
                expr = Concat(Value(prefix), F(model_field), Value(suffix))

        all_annotations[annotation_alias] = expr
        json_fields[field_name] = F(annotation_alias)

    # Final 'data' JSON structure
    all_annotations["data"] = JSONObject(**json_fields)

    self.log(f"All annotations: {all_annotations}")

    return use_read_replica_if_available(
        GeneratedCertificate.objects.filter(**filters)
        .annotate(**all_annotations)
        .order_by("created_date")
        .select_related("user")
        .values_list("data", flat=True)
    )

```
## Current State
We also reviewed the NAU PROD environment, and here are the certificates generated per month for the last two years:

```
2023-07: 4237 certificates  
2023-08: 7865 certificates  
2023-09: 7022 certificates  
2023-10: 10262 certificates  
2023-11: 11555 certificates  
2023-12: 15310 certificates  
2024-01: 9465 certificates  
2024-02: 14298 certificates  
2024-03: 17076 certificates  
2024-04: 12687 certificates  
2024-05: 34368 certificates  
2024-06: 14883 certificates  
2024-07: 7782 certificates  
2024-08: 6392 certificates  
2024-09: 8453 certificates  
2024-10: 11297 certificates  
2024-11: 12395 certificates  
2024-12: 12274 certificates  
2025-01: 9482 certificates  
2025-02: 11688 certificates  
2025-03: 14914 certificates  
2025-04: 15662 certificates  
2025-05: 18288 certificates  
2025-06: 15407 certificates  
2025-07: 6745 certificates  
```

We ran some local tests using 35k certificates (the max value in a month) with both approaches and got the following approximate metrics:

**First approach**
- Time: ~2 min
- Peak memory used: 7.89 MB

**New approach**
- Time: ~6 sec
- Peak memory used: 48.09 MB

The difference between the two approaches is significant. The new one is **95% faster** than the first, but still, 2 minutes doesn’t seem like a bad time to us. However, it’s important to note that the test was done with only one configured service, so a more realistic time would be:

- Approach Time x `N` (services)

At this point, we believe there are two possible paths forward:

1. **Proceed with the current approach** if the estimated execution time is acceptable. All unit tests are already implemented, so development could be completed sooner. We can run additional tests in the DEV environment before moving to PROD if necessary.

2. **Implement the optimized approach** now, minimizing in-memory iterations and leveraging the Django ORM to improve performance and scalability in the long term, particularly if the number of configured services increases. However, this option still requires validations to properly support computed fields, which may increase development time, especially considering that we will need to adapt the unit tests as well.


What do you think?

Tiago-Salles:
Hi @BryanttV and @efortish! Thank you for the considerations on improving the code performance. I will repeat the review regarding the fields that are not directly available. Have a good weekeng.

BryanttV:
Hi @Tiago-Salles!, we're implementing the new approach in this branch: [bav/send-course-certificates-engine](https://github.com/fccn/nau-openedx-extensions/tree/bav/send-course-certificates-engine), specifically, we include this new commit: https://github.com/fccn/nau-openedx-extensions/commit/6d63cfc570bbf6c8b47af1be3cfb0cd53cff587c. Can you check the changes?

nau_openedx_extensions/coursecertificate/engine.py
```diff
@@ -0,0 +1,257 @@
...
+                headers[auth_header] = f"Basic {credentials}"
+            elif auth_type == "api_key":
+                headers[auth_header] = auth_token
+
+        try:
+            response = requests.post(
+                api_url,
+                json=certificates_data,
+                headers=headers,
+                timeout=60,
```
igobranco: Timeout need to be parameterized.

nau_openedx_extensions/coursecertificate/engine.py
```diff
@@ -0,0 +1,257 @@
...
+
+            if certificates:
+                # Convert to service format
+                certificates_data = self.convert_certificates_to_service_format(certificates, service_config)
+
+                if certificates_data:
+                    # Send to service
+                    success = self.send_certificates_to_service(service_config, certificates_data, dry_run)
+                    if not success:
+                        self.log(f"Failed to send page {page_num} to {service_name}")
```
igobranco: ~Add missing break?~ 

Let it crash. Throw an exception.

igobranco: and throw an exception

nau_openedx_extensions/coursecertificate/engine.py
BryanttV: We could use a new attribute in the `config.yml` for each service:
```yml
endpoint_timeout: 60
```

BryanttV:
Hi @Tiago-Salles @igobranco. What do you think about the implementation of the new approach as mentioned [here](https://github.com/fccn/nau-openedx-extensions/pull/78#issuecomment-3070070965)? Do you have any suggestions? We'll be looking forward to knowing how to continue with the rest of the Milestones.

igobranco:
IMO we should use the 1st approach. 2 minutes is more than acceptable time for execution per remote integration. This estimation is what I were expecting from my experience, that is why I asked to use everything in memory and not asking you guys to go to Django ORM directly. 
I don't like the `model_field` approach because it will be harder for testing. It is too high level Python code and the configuration per service would need to know too much internal Python code structures.

igobranco:
Before merge this, we need to merge https://github.com/fccn/nau-openedx-extensions/pull/80

Tiago-Salles:
Hi @BryanttV and @efortish, I answered nothing as I was validating some concepts that I see they as valuable information. Thank you for considering building the second version, was great to learn new ways, and by sure they will be considered at the time this script requires some improvement. Since we already have an implementation improved, you guys made the time to process decrease considerably, and it also was implemented a new structure for the code, the first implementation becomes the main one. The second one sounds great, but it would requires from us a time to better fit the things, time that we don't have. Furthermore, I will leave here some considerations where you guys can check, but mainly to save in this ticket a way of improving the code in order to leave the encryption to MySQL. It should not be implemented since we don't have a way that indicates the nested model fields that should be also encrypted, and also considering the time of validation and implementation in our side, let's pass to the next steps.  


## Other possibilities

### Custom Django Func: `ToBase64`

This implements `Func` to execute `TO_BASE64` inside the MySQL engine for all the registers when necessary.

```python
from django.db.models import Func

class ToBase64(Func):
    function = 'TO_BASE64'
```

### `check_fields_to_encrypt` 

This is a new method that has the intention of recognize the transformations in order to build the expected query, leaving up to MySQL to return for all the fields a encrypted value.

```python
def check_fields_to_encrypt(self, service_config: dict):
    """
    For the given configuration:
 
    NAU_SEND_COURSE_CERTIFICATE_CONFIG:
    - service_name: portugal_digital
        endpoint_url: https://academiaportugaldigital.pt/api/api/Course/FinishedIntegrationNau
        auth_token: XXXXXXXXXXXXXXXXXXXXX
        auth_type: bearer
        auth_header: Authorization
        page_size: 100
        days: 30
        fields:
        - name: name
            func: nau_openedx_extensions.coursecertificate.extractors.student_email
            trans: md5
        - name: value
            func: nau_openedx_extensions.coursecertificate.extractors.course_id

    The return:

    {
        "name": {
            "encrypted_field_name": "name_MD5",
            "encryption_method": MD5(F(name))
        }
    }

    Or:

    {
        "name": {
            "encrypted_field_name": "name_BASE64",
            "encryption_method": ToBase64(F(name))
        }
    }
    """

    fields = service_config.get("fields")
    annotations = {}

    for field in fields:
        trans = field.get("trans")

        if trans:
            # We don't use `get` method here
            # because an invalid encryption
            # must fail the service
            try:
                trans_as_upper = trans.upper()
                encryption_method = {
                    "MD5": MD5(F(field["name"])),
                    "BASE64": ToBase64(F(field["name"]))
                }[trans_as_upper]
                encryption_field = f"{field['name']}_{trans_as_upper}"
                annotations[field['name']] = {
                    "encrypted_field_name": encryption_field,
                    "encryption_method": encryption_method
                }
            except Exception as e:
                raise e

    return annotations
```

### `process_service` 

The changes here are about to extract all the fields names, get the necessary transformations, and provide to the query method the fields and also the transformations.

```python
def process_service(self, service_config: dict, options: dict) -> None:
    """
    Process certificates for a specific service.
    This is the main entry point for the business logic.
    """
    ...

    service_fields = service_config.get("fields")
    service_fields = [f["name"] for f in service_fields]
    transformations = self.check_fields_to_encrypt(service_config)
    certificates_queryset = self.get_certificates_queryset(days, transformations, service_fields)

    # Apply service-specific filters
    certificates_queryset = self.apply_filters(certificates_queryset, service_config)

    ...
```

### get_certificates_queryset

The changes here are aimed to validate if it is necessary some transformation and apply it to the query. It removes the original field and replaces by the encrypted information. 

```python
def get_certificates_queryset(self, days: int, transformations: dict, service_fields: list = []) -> QuerySet:
    """Get certificates queryset filtered by date"""
    begin_date = datetime.now(UTC) - timedelta(days=days)
    query = GeneratedCertificate.objects.filter(created_date__gte=begin_date).order_by("created_date").select_related("user")
    
    if transformations:
        transformed_fields: list = []
        annotations = {}
        
        for original_name, trans_config in transformations.items():
            transformed_fields.append(trans_config["encrypted_field_name"])
            annotations[trans_config["encrypted_field_name"]] = trans_config["encryption_method"]
            service_fields.remove(original_name)
        
        fields = [*service_fields, *transformed_fields]

        return use_read_replica_if_available(
            query.annotate(**annotations).values(*fields)
        )
        
    return use_read_replica_if_available(query)
```

Tiago-Salles:
Since this is a new feature, it would be ideal to squash or reorganize the commits. This will make the commit history cleaner and easier to understand. Here is what I suggest:

- Reset the branch’s commits: ```git reset HEAD~<number_of_commits>```
- Stage and commit the changes incrementally, grouping them by logical features or steps.

Or if preferred they can be squashed.

BryanttV:
Hi @Tiago-Salles! 

We appreciate the time you dedicated and the valuable feedback you shared with us. We learned a lot during the process, especially by exploring new ways to optimize and structure the code. Without a doubt, it's something that can be very useful for future improvements. Thanks again!

Tiago-Salles:
> Hi @Tiago-Salles!
> 
> We appreciate the time you dedicated and the valuable feedback you shared with us. We learned a lot during the process, especially by exploring new ways to optimize and structure the code. Without a doubt, it's something that can be very useful for future improvements. Thanks again!

Hello @BryanttV, thank you for the feedback! I am glad it at least led us to something new.

I am validating the tests (which seem great) and after it, I will test the whole service by using it. As soon as I finish it, I let you know.

nau_openedx_extensions/coursecertificate/engine.py
Tiago-Salles: This point can applied doing an improvement in another moment. We are good to mark it as resolved.

Tiago-Salles:
After applying the change for this PR and for the tests, we are good to document and merge the first two PRs.

BryanttV:
Hi @Tiago-Salles, we are checking, but it is not clear which change you requested. :thinking:

Tiago-Salles:
> Hi @Tiago-Salles, we are checking, but it is not clear which change you requested. 🤔

Hi @BryanttV, I left a comment in a piece of code suggesting a change to be applied on `apply_transformations`. Could you confirm if you know what is this change I mean?

BryanttV:
@Tiago-Salles, are you referring to what was said here about applying the transformation using MySQL? (https://github.com/fccn/nau-openedx-extensions/pull/78#issuecomment-3079207635), or could you please point us to what suggestion you are referring to?

Tiago-Salles:
> @Tiago-Salles, are you referring to what was said here about applying the transformation using MySQL? ([#78 (comment)](https://github.com/fccn/nau-openedx-extensions/pull/78#issuecomment-3079207635)), or could you please point us to what suggestion you are referring to?

@BryanttV no, I will repeat the comment here, and delete the one created exactly above the piece of code. For some reason it seems you can't see, so I can be more practical.

In order to avoid breaking the execution because of typo, consider changing this method to something like:

```python
def apply_transformations(self, value: str, trans: str) -> str:
    """
    Apply transformations to a field value.
    Supports MD5 hash and Base64 encoding transformations.
    If the transformation type is not supported, returns the original value as string.
    Args:
        value (str): The value to transform.
        trans (str): The transformation type ('md5' or 'base64').
    Returns:
        str: The transformed value, or original value as string if transformation
            is not supported.
    """
    transformations = {
        "MD5": self._transform_md5,
        "BASE64": self._transform_base64,
    }
    
    if trans:
        try:
            transform_func = transformations.get(trans.lstrip().upper())
            return transform_func(value)
        except Exception as e:
            raise f"Encryption method not valid: {e}"
    return str(value)
``` 

And then things like `mD5`, `Md5`, `Base64`... Will not impact on it. Please apply these changes to tests.

**Note:** When a method is not supported it must fail the whole execution for this service, and not send any information.

Once we have this change done, and replicated to the tests we can merge the [tests](https://github.com/fccn/nau-openedx-extensions/pull/80) and proceed here. Please create tests also validating that different ways of typing the encryption method will not break the execution if the characters are correct. Thank you.

BryanttV:
@Tiago-Salles, thanks for your clarification and suggestion! I applied this change: https://github.com/fccn/nau-openedx-extensions/pull/78/commits/2e2c55e0a78692690faa33aa4e8020f9c8b89401

I also included new tests: https://github.com/fccn/nau-openedx-extensions/pull/80/commits/a294fa67c18b51f5c02b763e56aad3ab2f239498

Tiago-Salles:
> @Tiago-Salles, thanks for your clarification and suggestion! I applied this change: [2e2c55e](https://github.com/fccn/nau-openedx-extensions/commit/2e2c55e0a78692690faa33aa4e8020f9c8b89401)
> 
> I also included new tests: [a294fa6](https://github.com/fccn/nau-openedx-extensions/commit/a294fa67c18b51f5c02b763e56aad3ab2f239498)

Hi @BryanttV, thank you for applying the changes, the tests are good to be merged. Proceed there so I can approve here.

BryanttV:
Hi @Tiago-Salles, tests PR merged!

Tiago-Salles:
> Hi @Tiago-Salles, tests PR merged!

Great. Before merging here we need to merge [81](https://github.com/fccn/nau-openedx-extensions/pull/81) and [84](https://github.com/fccn/nau-openedx-extensions/pull/84).

BryanttV:
@Tiago-Salles, PR #81 and #84 merged!

BryanttV:
@Tiago-Salles, do you agree with merging this PR? Approval is still pending.

Or what do you think about merging https://github.com/fccn/nau-openedx-extensions/pull/85 and then merging this PR?

Tiago-Salles:
> @Tiago-Salles, do you agree with merging this PR? Approval is still pending.
> 
> Or what do you think about merging #85 and then merging this PR?

@BryanttV, let's merge here and change there to be targeted to master. How about?

BryanttV:
@Tiago-Salles Perfect, I agree! I'll proceed with the merge.

Tiago-Salles (approved):
LGTM!

PR #82, opened 2025-07-11
fix: validate dataresearcher role to export certificates

sandroscosta:
**Problem description**
Users with the data_researcher role weren't able to download certificate reports, unless they're part of the course team.
Only `data_researchers` and `super_admins` should be able to download course certificates.

**Solution**
I added the `CourseDataResearcherRole` to the list of roles to guarantee compliance, while removing the `Instructor` and  `Staff` roles

https://github.com/fccn/nau-technical/issues/615


sandroscosta:
Hey @BryanttV, can you review this, since you did the initial script?

BryanttV:
Hi @sandroscosta!

I have not yet tested the change, but I see that the `CourseDataResearcherRole` is not yet [here](https://github.com/fccn/nau-openedx-extensions/blob/95c3441489c7181a3686ce181a1a3ae5953a460b/nau_openedx_extensions/edxapp_wrapper/backends/student_l_v1.py#L7). Also we need to add the role in the [tests file](https://github.com/fccn/nau-openedx-extensions/blob/nau/redwood.master/nau_openedx_extensions/edxapp_wrapper/backends/student_l_v1_tests.py).

sandroscosta:
@BryanttV  I'll do that and fix the lint. No worries. I'll let you know once everything's finished.

sandroscosta:
@BryanttV done! When you're able to, please review. Thanks

BryanttV:
Thanks @sandroscosta, I'll be testing next week, although everything seems to be in order!

BryanttV (approved):
@sandroscosta, I tested in local and the permissions work correctly!

sandroscosta:
Thanks @BryanttV. I have to do some extra changes. After some internal discussing, we think that these commands shouldn't be available to team and staff. I'll remove staff privileges and leave only superadmin ones and dataresearcher.

sandroscosta:
@BryanttV can you recheck, please. Should be quick.

nau_openedx_extensions/edxapp_wrapper/backends/student_l_v1.py
```diff
@@ -4,7 +4,7 @@
 
 # pylint: disable=import-error, unused-import
 from common.djangoapps.student.models import CourseEnrollment, CourseEnrollmentAllowed
-from common.djangoapps.student.roles import CourseInstructorRole, CourseStaffRole
+from common.djangoapps.student.roles import CourseDataResearcherRole
```
BryanttV: Is it necessary to remove the `CourseInstructorRole` and `CourseStaffRole` roles? Perhaps it can be useful for future implementations. In case we want to remove it, we should also delete them in `student_l_v1_tests.py` and `student.py`.

BryanttV (changes requested):

nau_openedx_extensions/edxapp_wrapper/backends/student_l_v1.py
sandroscosta: I can leave them. I did that for the tests, so surely they can stay here for future usage.

BryanttV (approved):
Tested again locally, and everything works as expected!

sandroscosta:
@BryanttV Mega! Thank you so much for your time Bryan.

PR #84, opened 2025-07-16
feat: add event listener when an event is created

BryanttV:
## Description

This PR adds an event listener when the `CERTIFICATE_CREATED` event is sent in the platform. The command was also updated to receive a new parameter called `certificate-id`.

Each time a new certificate is created, in the background, a task is executed for each service configured in the `config.yml` file with the requested data of the new certificate.

Additionally, unit tests were included to validate the event listener, the new parameter in the command, and the engine logic.

## Related Issues
- https://github.com/fccn/nau-technical/issues/555

## Usage

You can use the new parameter by running the `send_course_certificates_by_web_service` command:

```bash
python manage.py lms send_certificates_by_web_service --certificate-id <certificate-id>
```
e.g:

```bash
python manage.py lms send_certificates_by_web_service --certificate-id 100
```

## Testing Instructions

1. Create a `tutor local` environment in the Redwood version. You need to run `tutor local` to use the workers.
2. Create a mount of `edx-platform` using the [NAU fork](https://github.com/fccn/edx-platform).
3. Install this plugin with the changes included in this pull request.
4. Create a course, configure grading, and certificates.
5. With a user, complete the course and request your certificate.
6. Automatically, the event listener runs in the background the command with the ID of the newly created certificate.
7. The certificate data is sent to all configured services if applicable.

### DIFF
--- a/nau_openedx_extensions/coursecertificate/handlers.py
+++ b/nau_openedx_extensions/coursecertificate/handlers.py
@@ -1,0 +1,52 @@
+"""
+Course certificate event handlers.
+"""
+
+import logging
+
+from django.core.management import call_command
+from django.dispatch import receiver
+from openedx_events.data import EventsMetadata
+from openedx_events.learning.data import CertificateData
+from openedx_events.learning.signals import CERTIFICATE_CREATED
+from openedx_events.tooling import OpenEdxPublicSignal
+
+from nau_openedx_extensions.edxapp_wrapper.certificates import GeneratedCertificate
+
+log = logging.getLogger(__name__)
+
+
+# pylint: disable=unused-argument
+@receiver(CERTIFICATE_CREATED)
+def certificate_created_send_to_external_services_handler(
+    signal: OpenEdxPublicSignal, sender, certificate: CertificateData, metadata: EventsMetadata, **kwargs
+):
+    """
+    Event handler that automatically sends newly created certificates to external services.
+
+    This handler is triggered when a certificate is created in the system. It retrieves
+    the corresponding GeneratedCertificate from the database and sends it to all
+    configured external services using the send_certificates_by_web_service command.
+
+    Args:
+        signal (OpenEdxPublicSignal): The signal that was sent.
+        sender: The sender of the signal.
+        certificate (CertificateData): The certificate data associated with the event.
+        metadata (EventsMetadata): The metadata of the event.
+        **kwargs: Additional keyword arguments.
+    """
+    user_id = certificate.user.id
+    course_id = certificate.course.course_key
+
+    try:
+        certificate_object = GeneratedCertificate.objects.get(user_id=user_id, course_id=course_id)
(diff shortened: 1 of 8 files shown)


nau_openedx_extensions/coursecertificate/handlers.py
```diff
@@ -0,0 +1,44 @@
...
+    Args:
+        signal (OpenEdxPublicSignal): The signal that was sent.
+        sender: The sender of the signal.
+        certificate (CertificateData): The certificate data associated with the event.
+        metadata (EventsMetadata): The metadata of the event.
+        **kwargs: Additional keyword arguments.
+    """
+    user_id = certificate.user.id
+    course_id = certificate.course.course_key
+    certificate_object = GeneratedCertificate.objects.get(user_id=user_id, course_id=course_id)
```
efortish, line 40: Can we add try-except block to handle possible errors? Something like:
```
try:
    certificate_object = GeneratedCertificate.objects.get(user_id=user_id, course_id=course_id)
except GeneratedCertificate.DoesNotExist:
    log.error(f"Certificate does not found with: user_id={user_id}, course_id={course_id}")
    return

try:
    call_command("send_certificates_by_web_service", certificate_id=certificate_object.id)
except Exception as e:
    log.exception(f"Error when trying to send certificate: {certificate_object.id} to external services: {e}")
```

efortish (changes requested):

nau_openedx_extensions/coursecertificate/handlers.py
```diff
@@ -0,0 +1,44 @@
...
+
+    log.info(f"Sending certificate {certificate_object.id} to external services")
+
+    call_command("send_certificates_by_web_service", certificate_id=certificate_object.id)
```
efortish: Can we add a try except block here to handle possible errors? Something like:
```
try:
        certificate_object = GeneratedCertificate.objects.get(user_id=user_id, course_id=course_id)
    except GeneratedCertificate.DoesNotExist:
        log.error(f"No GeneratedCertificate found for user_id={user_id} and course_id={course_id}")
        return
```

efortish: And:
```
try:
        call_command("send_certificates_by_web_service", certificate_id=certificate_object.id)
    except Exception as e:
        log.error(f"Failed to send certificate {certificate_object.id} to external services: {e}", exc_info=True)
```

efortish (changes requested):

nau_openedx_extensions/coursecertificate/handlers.py
BryanttV: Thanks! I added your suggestions here: https://github.com/fccn/nau-openedx-extensions/pull/84/commits/e503274f4afceec0e960c60b3b825af6211a0beb

efortish (approved):

Tiago-Salles (approved):
LGTM

PR #88, opened 2025-08-05
feat: allow skip course dirs

sandroscosta:
INA requested, for convenience, the possibility of skipping course directories for all reports are downloaded into the same directory. This is easier to load via Power BI.

* add a `--skip_course_dirs` arg to skip the creation of course directories.
* change the calculate_grades report type to grade_report, for it has parity with the name of the file downloaded.


Tiago-Salles (approved):
LGTM

PR #94, opened 2025-09-10
fix: render certificate for anonymous user

igobranco:
Fix the render course certificate when being viewed by an anonymous user.
Add a test for previewing the certificate.

This assumes that the CertificateRenderStarted Open edX filter has the user_certificate on its context.

fccn/nau-technical#667


Tiago-Salles (approved):

sandroscosta (approved):
LGTM

PR #95, opened 2025-10-29
feat: add is_active to the admin filters

sandroscosta:
Add an option to filter by active users to avoid sending newsletter to users that have not finished the account activation process.

fccn/nau-technical#702


Tiago-Salles (approved):
LGTM

PR #97, opened 2025-11-14
fix: add course staff role to certificate export roles

sandroscosta:
Teams were having issues when trying to extract certificate CSVs and PDFs due to incorrect permissions.

This PR adds course staff role to solve the issue raise by some teams about forbidden access.


Tiago-Salles (approved):
LGTM!

PR #117, opened 2026-01-14
feat: add org and course logo to ACE emails

igobranco:
Any email sent by ACE Open edX module would include organization and course logos (if available).

fccn/nau-technical#684


igobranco:
Replaced by #119 but instead of use Teak we will target Redwood.

PR #121, opened 2026-01-30
feat: monkey patch ACE

igobranco:
On ACE package monkey patch the `override_default_channel` per message type so we force to use our bulk channel instead of the ACE default one.

fccn/nau-technical#638


igobranco:
Replaced by:
- #122 for redwood

PR #129, opened 2026-02-20
fix: certificate PDF download URL formatting

igobranco:
The certificate download PDF link was missing a forward slash separator between the base URL and the certificate UUID, causing malformed URLs like "https://example.com/certificatestest-uuid-123" instead of the correct "https://example.com/certificates/test-uuid-123".

fccn/nau-technical#795


Tiago-Salles (approved):
LGTM

PR #136, opened 2026-03-11
fix: update requirements

igobranco:
Fix the requirements workflow on Makefile target to include pre-requirements.


ManuelStarDo (approved):
lgtm

PR #138, opened 2026-03-13
feat: check advanced settings to apply the filter

Tiago-Salles:
Related to: fccn/nau-technical#800


igobranco (commented):
@Tiago-Salles  this exception needs to be translatable.

https://github.com/fccn/nau-openedx-extensions/blob/fb178c7c137955e146320b147b14d9bea54b5b8a/nau_openedx_extensions/partner_integration/course_filters.py#L94-L95

PR #141, opened 2026-03-18
chore: remove the SSO constraint

Tiago-Salles:
In an internal conversation it was pointed out the SSO must to be flexible in terms of constraints to be able to handle the different scenarious it must to carry out.


Tiago-Salles:
@ManuelStarDo, this changes the behavior we found on our tests.

ManuelStarDo (approved):
LGTM!

PR #145, opened 2026-04-30
Block CourseEnrollmentAllowed when filter_enrollment_require_nif is e…

ccantillo:
Related issue: fccn/nau-technical#771

#  `CourseEnrollmentAllowed` safeguard vs `filter_enrollment_require_nif`

## Problem

 when a course has **`filter_enrollment_require_nif`** enabled in *other course settings*, teams could still create **Course enrollment allowed** invitations (`CourseEnrollmentAllowed`). When the learner activates their account, the platform tries to auto-enrol them; the **`FilterEnrollmentRequireNIF`** pipeline step rejects enrolment if there is no NIF / Autenticação Gov on the account. That surfaces as a **500 in Django Admin** (and learners who cannot activate normally), as reported in support (e.g. NSU-37572 and related tickets).

Root cause: an **incompatible state** — an allowed invitation exists for a course that requires NIF before enrolment can succeed.

## Solution

- Add **`course_enrollment_allowed_guard.py`**, which validates on each **`pre_save`** of `CourseEnrollmentAllowed` (skipped when `raw=True`).
- If the course has **`filter_enrollment_require_nif`** enabled (same read path as `FilterEnrollmentRequireNIF` via `get_other_course_settings`), raise **`ValidationError`** with a clear staff-facing message so that invitation cannot be created or saved.
- Register the receiver from **`signals.py`** via **`connect_course_enrollment_allowed_nif_guard()`**, with **lazy import** of the model through **`get_course_enrollment_allowed_model()`** in **`edxapp_wrapper/student.py`** so the guard module does not touch `settings` at import time.
- The test stub **`student_l_v1_tests`** sets **`CourseEnrollmentAllowed = None`** so the signal is not wired when edx-platform is not loaded.


efortish:
@ccantillo hello:

Please in the PR description add at the top the related issue.

Something like: Related issue: https://github.com/fccn/nau-technical/issues/771

igobranco (approved):
Looking rapidly for the code, it seems to be ok.
Let's merge it.

PR #147, opened 2026-06-25
fix: adjusted certificate filter query

Tiago-Salles:
This adjusts the filter query on the webservice

Related to: fccn/nau-technical#904


ManuelStarDo (approved):
LGTM ✔️

PR #148, opened 2026-06-25
fix: adjust enrollments filter query

Tiago-Salles:
This adjusts the filter query on the webservice

Related to: fccn/nau-technical#905


ManuelStarDo (approved):
### Merge readiness
**Recommendation:** Good to merge.  
The change is coherent, safe, and appropriately tested.

### Suggested improvements
- Use `self.assertFalse(result.exists())` or `self.assertEqual(result.count(), 0)` for stronger Django-style test clarity.
- If desired, rename the new test slightly to emphasize preservation of prior filtering, not just scope application.

PR #149, opened 2026-07-13
feat: added management view for SSO

Tiago-Salles:
This PR introduces a new view for managing SSO operations in the Partner Integration module. The `PartnerSSOManagementView` class provides two main functionalities:

### Delete SSO Register

A `delete` method implementation that allows clients to remove an SSO register by providing the `external_user_id`. If the register exists, it will be deleted, and a success response will be returned. If the register does not exist, a 404 error will be returned.

### Retrieve SSO Register

A `get` method implementation allows clients to retrieve an SSO register associated with the authenticated partner client by providing the `external_user_id`. If the register exists, it will return the serialized data. If the register does not exist, a 404 error will be returned. If the `external_user_id` is not provided, a 400 error will be returned.

It is important to note that the current `get` (retrieve) implementation requires the `external_user_id` to retrieve SSO register, that is, it only filters. However, future enhancements may allow for retrieving all registers for a given client by considering additional parameters beyond just the `external_user_id`.

Example of response:

```json
{
    "external_user_id": "12345",
    "username": "john_doe",
    "email": "john.doe@example.com"
}
```

Receiving a response containing the SSO register means that the link between the external user and the internal user is valid and active. Otherwise, it would never find any matching register.

> The view was created focused on the current necessity, but it is open to growth to support more complex queries and operations related to SSO management.

Related to: fccn/nau-technical#956

### DIFF
--- a/nau_openedx_extensions/partner_integration/serializers.py
+++ b/nau_openedx_extensions/partner_integration/serializers.py
@@ -240,3 +240,10 @@ class CourseProgressSerializer(serializers.Serializer):
     completion_summary = CompletionSummarySerializer()
     course_grade = CourseGradeSerializer()
     grading_policy = GradingPolicySerializer()
+
+
+class SSOUserSerializer(serializers.Serializer):
+    """Serializer for SSO user data in data extractor responses."""
+    external_user_id = serializers.CharField()
+    username = serializers.CharField(source='user.username')
+    email = serializers.EmailField(source='user.email')
--- a/nau_openedx_extensions/partner_integration/tests/test_sso.py
+++ b/nau_openedx_extensions/partner_integration/tests/test_sso.py
@@ -190,3 +190,173 @@ class TestCustomAuthorizationView(TransactionTestCase):
 
         self.assertEqual(response.status_code, 302)
         self.assertEqual(response.url, "https://www.nau.edu.pt")
+
+
+class TestPartnerSSOManagementView(TransactionTestCase):
+
+    def setUp(self):
+        self.http_client = APIClient()
+        self.endpoint = "/nau-openedx-extensions/partner-integration/sso/manage/"
+
+        self.partner_client = PartnerAPIClientFactory.create(
+            is_active=True,
+            query_security_scope={"base_security_scope": {"org": f"TEST_ORG"}}
+        )
+        self.partner_client.password = "correct_password"
+        self.partner_client.save()
+
+        self.jwt_token = self.authenticate_partner_client(self.partner_client)
+
+    def authenticate_partner_client(self, partner_client):
+        """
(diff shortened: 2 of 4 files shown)


ManuelStarDo (commented):
## Review
### Good points
- The feature is clearly scoped and useful for operational support.
- Authentication and partner-client permission classes are already applied.
- Error handling for missing `external_user_id` and missing records is present.
- Test coverage is fairly strong for the main happy paths and common failure cases.

### Concerns
#### 1. Broad exception handling may hide real API problems
Both `get` and `delete` catch generic `Exception` and return `500` with `str(e)` to the client.

Why this matters:
- It may leak internal implementation details.
- It makes debugging and client behavior less consistent.
- It can unintentionally expose sensitive information.

Better approach:
- Return a generic internal error message to the client.
- Keep detailed exception info only in logs.

#### 2. Logging message in `delete` says POST instead of DELETE
Inside `delete()`:
- `logger.info("PartnerSSOManagementView: POST request received.")`

This is clearly incorrect and should be fixed.

#### 3. Unused variable in `delete`
Inside `delete()`:
- `query_security_scope: dict = client.query_security_scope`

This variable is assigned and never used.

Why this matters:
- It adds noise.
- It suggests incomplete implementation or dead code.

#### 4. 204 response body usage
The delete method returns:
- `Response({"success": True}, status=204)`

Technically, HTTP 204 means no content. A body should generally not be returned.

Better approach:
- Return `Response(status=204)`
- or switch to `200` if you want to return JSON confirmation.

#### 5. Minor wording / doc consistency issues
The docstrings and comments use inconsistent wording:
- “manage SSO operations”
- “retrieve SSO registers”
- singular vs plural response behavior

This is not blocking, but refining naming would make the endpoint easier to maintain.

### Merge readiness
**Recommendation:** Mergeable after small corrections.  
The feature is good and the overall implementation is solid, but I would prefer a few cleanups before merge because they directly affect API quality and maintainability.

**Status:** Conditionally good to merge, but better if corrected first.

### Suggested improvements
- Fix the incorrect log message in `delete()` from `POST` to `DELETE`.
- Remove the unused `query_security_scope` variable.
- Do not return a response body with HTTP 204.
- Avoid returning raw exception text to API clients; log it internally and return a generic 500 message.
- Tighten the docstrings to consistently describe singular lookup behavior.

ManuelStarDo (approved):
Thanks, this looks much better now. 

That said, because this is an identity-management endpoint, I’d still strongly recommend adding explicit regression tests for partner-client isolation — i.e. proving that one authenticated partner client cannot retrieve or delete another client’s SSO record even if it knows the same external_user_id.

If you want one more small improvement, I’d also add tests for the inactive client path on both GET and DELETE.

This is non blocking though, so i am approving and merging the PR.

PR #150, opened 2026-07-20
feat: enhance SSOPartnerIntegration admin for support troubleshooting

Tiago-Salles:
## Changes applied

- Add user-centric search in SSOPartnerIntegration admin using username/email/id
- Display partner client, external_user_id, and email in list view
- Show created_at and updated_at timestamps for association records
- Use raw_id_fields for user lookups and list_select_related for performance

## Admin pages

image

image

Related to: https://github.com/fccn/nau-technical/issues/958


ManuelStarDo (commented):
### Review
### Good points
- This is a useful operational/admin enhancement.
- Search improvements are practical for support workflows.
- `list_select_related` and `raw_id_fields` are sensible additions.
- Adding timestamps improves auditability.
- Test coverage for admin configuration is a plus.

### Concerns
#### 1. Broad exception in `openedx_email`
```python
def openedx_email(self, instance):
    try:
        return instance.user.email
    except Exception as error:
        return str(error)
```

Why this matters:
- Broad exception handling in admin display methods is unnecessary here.
- Returning raw exception text in admin UI can create confusing output.
- If `user` is required and properly related, this should normally not fail.

Better approach:
- Return a safe fallback such as `""` or `None`
- or just access `instance.user.email` directly if the relation is guaranteed

#### 2. `user__id` in `search_fields`
Django admin search is generally intended for text-like lookups. Including `user__id` may work less predictably depending on how admin builds the search query and database casting.

Why this matters:
- Numeric ID search in admin often deserves a custom `get_search_results` implementation instead of relying only on `search_fields`.

That said, this may still work in your environment, so I see this more as an improvement point than a blocker.

#### 3. Read replica in admin search
Overriding `get_search_results()` to wrap the queryset with `use_read_replica_if_available(qs)` is interesting, but it is worth validating whether admin search consistency is acceptable on a replica.

Why this matters:
- Operational support sometimes expects very fresh writes to be immediately searchable.
- Replica lag could make troubleshooting confusing.

Not necessarily wrong, but worth confirming.

#### 4. Migration field definition detail
The migration sets:
- `created_at = auto_now_add=True, null=True, default=None`
- `updated_at = auto_now=True, default=timezone.now`

This is probably fine for backfilling existing rows, but the model definition and migration intent could be cleaner if defaults/nullability are aligned deliberately and documented.

### Merge readiness
**Recommendation:** Good to merge, with minor follow-up improvements.  
No blocking issue stands out, and the enhancement is useful. Most concerns are quality refinements rather than correctness problems.

### Suggested improvements
- Replace broad exception handling in `openedx_email()` with a simpler and safer fallback.
- Validate whether `user__id` search behaves as intended in admin.
- Confirm that read-replica search is acceptable for support/admin use cases.
- Optionally revisit the timestamp migration defaults for clarity.


@Tiago-Salles 
Id like to hear your thoughts on these points first before approval, specially on the `user__id` validation and read-replica search function

ManuelStarDo (approved):
Thanks for the clarifications — with the confirmation that user__id search works as expected and that read-replica usage here is intentional for this admin flow, I don’t see a blocker anymore. The migration rationale for existing records also makes sense.
I’d treat the openedx_email() exception handling as a minor cleanup opportunity, not a merge blocker. 
From my side, this looks good to approve.

PR #151, opened 2026-07-21
fix: pin Django version on tests

Tiago-Salles:
This was required to fix the lint command that counts on resources from upstream that are only compatible with Django versions `<5.0`. 

Related to: fccn/nau-technical#963


ManuelStarDo (approved):
LGTM ✔️

PR #152, opened 2026-07-22
fix: adjusted student progress facade filters

Tiago-Salles:
Adjusted filtering logic.

Related to:
- fccn/nau-technical#902
- fccn/nau-technical#903

### DIFF
--- a/nau_openedx_extensions/partner_integration/facade.py
+++ b/nau_openedx_extensions/partner_integration/facade.py
@@ -452,9 +452,9 @@ class StudentProgressExportFacade(DataExtractorFacade):
             if student_id:
                 return User.objects.get(id=student_id)
             elif nif:
-                return User.objects.get(nauuserextendedmodel__nif=student_id)
+                return User.objects.get(nauuserextendedmodel__nif=nif)
             elif username:
-                return User.objects.get(nauuserextendedmodel__username=username)
+                return User.objects.get(username=username)
 
             return User.objects.get(email=email)
         except Exception as e:


ManuelStarDo (commented):
### Review
### Good points
- The bug fix is straightforward and clearly correct by inspection.
- Very low implementation risk.
- It directly addresses obviously broken filter paths.

### Concerns
#### 1. Missing regression tests
This is the main issue in the PR.

The fix changes lookup behavior in an important part of student filtering logic, but no test was added. Since both previous lines were wrong, this is exactly the kind of bug that should be protected by tests.

Suggested coverage:
- lookup by `nif`
- lookup by `username`
- precedence behavior if multiple identifiers are provided

### Merge readiness
**Recommendation:** Not ideal to merge yet.  
The code fix itself looks correct, but I would request tests before approving because this is a regression-prone area and the PR currently has no proof of behavior.

### Suggested improvements
- Add regression tests for `nif` and `username` lookups.
- Consider adding a test for precedence order among `student_id`, `nif`, `username`, and `email`.
- If this method is user-facing through an API flow, ensure invalid/duplicate match behavior is also covered.


@Tiago-Salles 
Can you make some tests for this?

ManuelStarDo (changes requested):
@Tiago-Salles 
I re-reviewed the code, we can make and follow the tests on another issue/PR

However, there is an actual blocking issue in the code regarding the username fix — technically works, but introduces a new bug. The original line (nauuserextendedmodel__username=username) was broken because NauUserExtendedModel has no username field — it would raise a FieldError on every call. The new line traverses nauuserextendedmodel__user__username, which does resolve to a valid field path (going back through the one-to-one relation to User.username), so it will no longer error out.

However, this traversal performs an implicit inner join against NauUserExtendedModel. Since NauUserExtendedModel.user is null=True and not every User is guaranteed to have a related NauUserExtendedModel row (e.g., a user who never went through the CC/NIF registration flow — see third_party_auth/pipeline.py where the model is only created conditionally on ensure_cartao_cidadao_data), this query will fail to find users who exist but don't have an extended-model record, even though a simple username lookup should trivially succeed for them.

Compare with the existing, correct pattern used just above in EnrollmentFacade.enroll_user (same file, around line 289):

    elif username and not user:
        user = use_read_replica_if_available(User.objects.filter(username=username)).get()

That's a direct User.objects.get(username=username) — no unnecessary join. The fix in this PR should follow the same pattern:

    return User.objects.get(username=username)

Tiago-Salles:
Hey @ManuelStarDo, thanks for this last review, I have applied the remaining change.

ManuelStarDo (approved):
##  Possible Improvements

- **[MEDIUM]** No test currently exercises the `nif` or `username` branches of `_get_student_user` (only `student_id` and `email` are covered in `test_facade.py`, lines 481–494). Given this method was bugged in *two* separate branches simultaneously and shipped undetected, adding two small unit tests (mirroring the existing `test_get_student_user_by_student_id`/`test_get_student_user_by_email` pattern) would have caught both regressions and would prevent silent recurrence. This is a low-cost, high-value addition given the method already has a `TestStudentProgressExportFacade` test class in place to extend. — `nau_openedx_extensions/partner_integration/tests/test_facade.py:481`

- **[LOW / SUGGESTION]** Inconsistency with `EnrollmentFacade.enroll_user`: the enrollment facade's NIF lookup checks both `nauuserextendedmodel__nif` **and** `nauuserextendedmodel__cc_nif` (line 283–286 of `facade.py`), but `StudentProgressExportFacade._get_student_user` only checks `nauuserextendedmodel__nif`. If `cc_nif` is meant to be an alternative valid identifier across the codebase (it's used for NIF-based certificate/enrollment filtering too), this facade may be missing coverage for citizen-card-issued NIFs. Worth confirming with the team whether this was an intentional scope reduction for student progress lookups or should be aligned for consistency. — `nau_openedx_extensions/partner_integration/facade.py:455`

- **[LOW / SUGGESTION]** The broad `except Exception as e:` in `_get_student_user` (line 460) swallows the specific `FieldError` that the old buggy code would have raised, always converting it into a generic `PartnerIntegrationInternalErrorException`. This masked the bug from callers/logs beyond the generic log line. Not a blocker (this pattern is used consistently throughout the file), but worth being aware that logic errors like this one can hide behind broad exception handling — consider adding assertions or narrower except clauses in critical lookup paths if better fault visibility is desired.


@Tiago-Salles 
We are good to merge this, thank you for the work.
I must once again reinforce the request to implement tests for all these functionalities to prevent regressions and improve maintainability in the future

Tiago-Salles:
Hey @ManuelStarDo! We are good to merge, thanks.

PR #153, opened 2026-07-23
fix: recover choice text for div-wrapped MCQ/checkbox choices in problem responses report

rguerra-fccn:
## What
Monkeypatches `ChoiceGroup.extract_choices` and `LoncapaProblem.find_correct_answer_text` so the "Problem Responses" instructor report correctly resolves answer text for multiple-choice/checkbox problems whose choices were authored with Studio's rich-text / per-choice feedback editor.

## Why
Fixes 
- fccn/nau-technical#948

That editor wraps each choice's visible text in a child element (e.g. `<div>`) instead of storing it as a direct text node of `<choice>`. The report-generation code only reads the immediate text node (`choice.text` / the `text()` XPath axis), so it falls back to `"Answer Text Missing"` / an empty correct-answer value for these choices — even though rendering and grading work fine for students, since those code paths recurse into child elements (`stringify_children`).

## Fix
Both lookups now fall back to a recursive text extraction (skipping `<choicehint>`/`<compoundhint>` feedback children) only when the original implementation returns an empty result, leaving all other behaviour unchanged.

This is a stopgap fix for an upstream Open edX bug. An upstream ticket will be filed once we've validated this works across environments, referencing this monkeypatch so it can be removed once fixed upstream.

## Testing
- Added unit tests covering both factories (happy path, passthrough, div-wrapped recovery, edge cases) — 8/8 passing.
- Reproduced the bug against real production OLX and confirmed the fix resolves it, with no regressions for plain-text (non-div) choices.
- Validated end-to-end in local dev environment: created a throwaway test course with the real broken OLX, answered it, and confirmed the generated "Problem Responses" CSV now shows the correct answer text via the instructor dashboard.

### DIFF
--- a/nau_openedx_extensions/apps.py
+++ b/nau_openedx_extensions/apps.py
@@ -65,3 +65,25 @@ class NauOpenEdxConfig(AppConfig):
             get_course_email_context_factory  # pylint: disable=import-outside-toplevel # noqa
         bulk_email_tasks._get_course_email_context = \
             get_course_email_context_factory(prev_email_context_func)  # pylint: disable=protected-access
+
+        # Fix the "Problem Responses" report showing "Answer Text Missing" / an empty
+        # correct answer for multiple-choice/checkbox problems whose choices were authored
+        # with Studio's rich-text / per-choice feedback editor (see nau-technical#948).
+        # TODO(nau-technical#948): remove once fixed upstream in edx-platform/xblocks-contrib.
+        from xmodule.capa.capa_problem import \
+            LoncapaProblem  # pylint: disable=import-error,import-outside-toplevel # noqa
+        from xmodule.capa.inputtypes import \
+            ChoiceGroup  # pylint: disable=import-error,import-outside-toplevel # noqa
+        from nau_openedx_extensions.xblocks.capa_problem_responses import (  # pylint: disable=import-outside-toplevel # noqa
+            get_extract_choices_factory,
+            get_find_correct_answer_text_factory,
+        )
+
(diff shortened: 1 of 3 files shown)


ManuelStarDo (approved):
**Review: Approved with minor suggestions**

Nice, well-documented fix — the root-cause analysis (rich-text choice editor wrapping text in `<div>` vs. the report code only reading direct text nodes) is clear, and the fallback-only-on-empty-result pattern is a safe way to patch `ChoiceGroup.extract_choices` / `LoncapaProblem.find_correct_answer_text` without touching existing behavior. 
Good test coverage (happy path, passthrough, div-recovery, `optioninput` fallback, edge cases) plus manual validation against real broken OLX in a throwaway course — that's solid due diligence for a monkeypatch like this.

Two minor suggestions, non-blocking:
1. In `find_correct_answer_text_wrapper`, the XPath is built via string concatenation of `answer_id` (`'//*[@id="' + answer_id + '"]'`). Even though `answer_id` should always be internally generated, consider using an XPath variable binding instead, to avoid any fragility if it ever contains a `"`.
2. In `extract_choices_wrapper`, `zip(choices, choice_elements)` silently truncates if the two lists differ in length. Might be worth an assertion/log if they diverge, just so any future upstream change to `extract_choices`'s return shape fails loudly instead of silently misaligning choice text.

ManuelStarDo (approved):
Perfect @rguerra-fccn 

LGTM ✔️ 

SInce you tested this locally, we can merge, deploy to DEV and test there

PR #155, opened 2026-08-11
SSO account linking flow

Tiago-Salles:
## What this changes

An SSO link between a NAU account and a partner identity could be reassigned by a browser request, and an open NAU session was enough to decide which account a partner identity belonged to. On a shared computer — a previous user leaves their NAU session open — that silently moved a partner identity onto the wrong account.

This makes the link itself the authority on who a request belongs to, and moves the intentional change of an `external_user_id` to an authenticated server to server operation.

## Behaviour

**Authorization flow (`CustomAuthorizationView`)**

- A link is only ever **created** here, never reassigned. Reassigning the `external_user_id` of an existing link, or claiming one that already belongs to another NAU user, is refused.
- Refusals redirect to the application's registered redirect URI with `?error=sso_link_conflict`, so the partner can tell the user what happened rather than the error landing on a NAU page the partner never sees.
- When the browser session belongs to someone other than the link owner, the session is dropped and the flow restarts, so the partner is authorized against the owner of the identity instead of a leftover session.

**Management endpoint (`PartnerSSOManagementView`)**

- New `PATCH`: changes the `external_user_id` of an existing link. Authenticated by the partner client, addressed by an identifier the partner already holds, so it never depends on which session is open in a browser. Returns `409` when the new identifier is already linked to another NAU user. The NAU user of a link is never changed here — moving a link means deleting it and linking again.
- `GET`, `DELETE` and `PATCH` now accept a NAU `username` as an alternative to `external_user_id`, which keeps these operations usable when the partner no longer holds the identifier currently stored on the NAU side.

## Why the constraint rather than a check before the insert

Uniqueness of `external_user_id` on creation cannot be decided by a read. The creation path runs precisely when no register held that identifier, so a read there is answered by the state that sent us down the path, and it cannot see a concurrent request claiming the same identifier in between. The unique constraint can, and `IntegrityError` is translated into the same `sso_link_conflict` result.

The equivalent check on `PATCH` is kept, because it *is* reachable there and gives a precise message; the constraint covers the read-to-save window behind it.

Related to: https://github.com/fccn/nau-technical/issues/1013


rguerra-fccn (approved):
**Approving after live functional testing on a real dev environment** (checked out the branch, applied the migration against a real MySQL dev DB, and exercised the SSO flow end-to-end via genuine HTTP requests — real login, real cookies, real `curl`, no mocking).

**Scenarios verified:**
- ✅ First-time SSO link creation — correct redirect, DB row created as expected
- ✅ Same user attempting to link a *different* `external_user_id` — correctly refused with `error=sso_link_conflict`, no reassignment
- ✅ **Shared-computer / stale-session case** (the core fix): a browser session belonging to one NAU user hitting an SSO link owned by another user — confirmed via server-side session decoding that the stale session is dropped and the flow restarts as the *actual* link owner, never silently reusing the wrong session
- ✅ **Genuine concurrent race**: two real simultaneous requests claiming the same `external_user_id`, on both the `authorize` GET path and the `PartnerSSOManagementView.patch` path — one wins cleanly, the other gets a real `IntegrityError` → clean `409`/conflict response. No transaction errors in the logs, no duplicate rows, connection stayed healthy for subsequent requests.

Migration is additive (stricter uniqueness constraints only) and applied cleanly to a live DB with existing data.

Nice fix for a real security gap — LGTM 🚀

ManuelStarDo (changes requested):
## The following concrete issue is critical to address before merging:

- **`nau_openedx_extensions/partner_integration/course_filters.py` — `FilterSSOPartnerAccountLink.run_filter` breaks under the new `user` cardinality.**
  It calls `SSOPartnerIntegration.objects.get(user=user)` (single-argument lookup, no `partner_client` filter). Before this PR, `user` was a `OneToOneField`, so at most one row could ever exist per user and this call was safe. After this PR, `user` is a plain `ForeignKey` with uniqueness only on `(user, partner_client)` — a user can now legitimately hold multiple `SSOPartnerIntegration` rows (one per partner), which is in fact one of this PR's intended improvements (see `test_sso_creates_a_link_for_a_user_linked_to_another_client`). As soon as any user has two or more partner links, `SSOPartnerIntegration.objects.get(user=user)` will raise `MultipleObjectsReturned`, and since `course_filters.py` only catches `SSOPartnerIntegration.DoesNotExist`, this exception will propagate uncaught and (depending on `fail_silently` in `OPEN_EDX_FILTERS_CONFIG`) can outright block enrollment for that course with a 500-style failure instead of the intended `PreventEnrollment` UX.
- Risk: Silent breakage of an existing, unrelated enrollment-filter feature for any user who links to more than one partner — exactly the scenario this PR is designed to newly allow.
- Fix: Scope the lookup by the course's associated partner (or by all matching rows, picking/validating the relevant one), e.g. filter by the partner client that owns the course's security scope, or iterate `SSOPartnerIntegration.objects.filter(user=user)` and pick the one whose `partner_client` can access `course_key`. At minimum, catch `MultipleObjectsReturned` alongside `DoesNotExist` so the existing "no linked partner" UX degrades gracefully rather than raising unhandled.

## Possible Improvements

- **`views.py:540` — the session-drop/restart path calls `redirect(request.get_full_path())` without any loop guard.** If, for some reason, the freshly-restarted request still resolves to a session belonging to a different user than the register owner (e.g., a caching proxy, a browser auto-restoring a cookie, or another tab racing the logout), this could redirect indefinitely. In practice `logout()` should prevent this, but consider a one-shot marker (e.g., a query param or session flag) to make the restart provably terminate rather than relying solely on `logout()` succeeding.
- **`views.py:606` (`handle_sso_link_conflict`) silently falls back to `DEFAULT_PARTNER_SSO_REDIRECT_URI` when `application.redirect_uris` is empty**, sending the user to a generic NAU page with `?error=sso_link_conflict` appended — but the NAU landing page presumably won't parse or display that query param, so the "partner can inform the user" goal silently fails for partners with no configured `redirect_uris`. Consider logging a warning in this branch so misconfigured partner apps are discoverable operationally.
- **`README.md`, last line: "It corrects automatically a register that has changed its main information (`external_user_id`)."** This sentence is a leftover from the pre-PR `DELETE` documentation and no longer matches `DELETE`'s actual (and now correct) behavior — deletion doesn't "correct" anything, it removes the register outright, and the "correcting" behavior it originally described was the overwrite behavior this PR just removed. Worth deleting this stray line to avoid confusing future readers.
- **`handle_sso_registration`'s docstring/log messages use f-strings with `partner_client.name`**, which is fine, but note the pre-existing broad `except Exception as e: raise e` at the end of the method (unchanged by this PR) still re-raises everything uncaught by the two explicit paths, ultimately surfacing as the view's generic 500 in `dispatch`'s outer `except Exception`. Not a regression, but since this PR is already hardening this exact method, it would be a good opportunity to also catch `Application.DoesNotExist` explicitly here (currently only handled in `dispatch`, not in `handle_sso_registration`, even though `Application.objects.get(client_id=sso_client_id)` is called directly inside it now at line 573) — today an invalid `sso_client_id` reaching this method could raise unhandled instead of hitting `DEFAULT_PARTNER_SSO_REDIRECT_URI` gracefully.

rguerra-fccn (changes requested):
Agreed with the `course_filters.py` finding — verified it independently. `FilterSSOPartnerAccountLink.run_filter()`'s `SSOPartnerIntegration.objects.get(user=user)` was safe under the old `OneToOneField`, but the new `ForeignKey` + `(user, partner_client)` uniqueness lets a user hold multiple links (confirmed intended by this PR's own test), so that `.get()` now raises `MultipleObjectsReturned` — uncaught, since only `DoesNotExist` is handled. This filter is live in production/dev with `fail_silently: False`, so it's reachable, not theoretical.

I swept the rest of the repo for the same bug class (any other consumer assuming one-row-per-user) — this is the only one; `views.py`/`admin.py` all scope lookups by `partner_client` too, still guaranteed unique.

Fix: scope by the relevant partner client, and/or catch `MultipleObjectsReturned` alongside `DoesNotExist`.

On the improvements list: the missing `Application.DoesNotExist` catch in `handle_sso_registration` is worth fixing alongside the blocker — it's called outside `dispatch`'s try/except, so an invalid `client_id` there would surface as an unhandled 500 instead of the graceful redirect. The redirect-loop guard, silent-fallback logging, and stray README line are fine as optional polish, not blocking.

Tiago-Salles:
Hey @ManuelStarDo, I have applied the changes, ready to review.

ManuelStarDo (approved):
## Merge Ready ✔️ 
@Tiago-Salles Thank you for the work!
This PR looks good, I will merge it now.

## Minor improvements

**[LOW] `nau_openedx_extensions/partner_integration/views.py:596-608` — IntegrityError message can mislabel which constraint fired**
Risk: `handle_sso_registration` only catches `IntegrityError` around the `create()` call and always attributes it to the `(partner_client, external_user_id)` constraint ("external user ID is already linked to another NAU user"). But a concurrent request for the *same user* being linked to the *same partner* for the first time (two tabs, two different `external_user_id`s racing before either commits) would violate the `(user, partner_client)` constraint instead, and get the same misleading log message. The user-facing behavior (redirect to `sso_link_conflict`) is still correct either way, so this is a diagnostics-only issue, not a functional bug.
Fix: Either inspect the constraint name from the `IntegrityError` (Postgres) or re-query for an existing `(user, partner_client)` row after the exception to log an accurate message distinguishing "user already linked to this partner" vs "external id already claimed."

**[LOW / SUGGESTION] `nau_openedx_extensions/partner_integration/views.py:566-619` — Redundant re-validation of partner client/application in `handle_sso_registration`**
Risk: None functionally — just duplicated work. `dispatch()` already validates the JWT and fetches the `Application` before calling `handle_sso_registration`, which then re-validates the JWT (`ClientJWTAuthentication().validate_token_data_and_return_client`) and re-fetches the `Application` by `client_id`. Harmless, but it's an extra DB/JWT-decode round trip on every registration path.
Fix: Consider passing the already-resolved `partner_client`/`application` objects into `handle_sso_registration` instead of re-deriving them from raw `jwt_token`/`sso_client_id`.

**[LOW / SUGGESTION] Missing type hints on new public methods**
`build_register_lookup`, `handle_sso_registration`, `handle_sso_link_conflict`, `_build_session_restart_url` are all new/rewritten public-ish methods without type hints on parameters or return values (the rest of the file is inconsistent on this too, so it's not a regression, but worth tightening given how much of this method surface changed).
Fix: Add hints, e.g. `def build_register_lookup(client: PartnerAPIClient, external_user_id: str | None = None, username: str | None = None) -> tuple[dict | None, str | None]:`.

PR #156, opened 2026-08-15
docs: add ADR 0001 on report normalization

efortish:
Records the Phase 2 decision on what normalizing course reports means: a base column contract defined per row grain, enforced by wrapping the shared instructor_task write path from the plugin, since instructor_task exposes no openedx-filters extension point and edx-platform is not to be modified.

Refs fccn/nau-technical#955

### DIFF
--- a/docs/decisions/0001-report-normalization.rst
+++ b/docs/decisions/0001-report-normalization.rst
@@ -1,0 +1,295 @@
+0001: Common Base Structure for All Course Reports
+###############################################
+
+Status
+******
+
+**Draft** *2026-08-14*
+
+Scope
+*****
+
+Phase 2 (DRC) only, and within it only what report normalization means and how
+it is implemented.
+
+- **Here:** the base column contract and where it is enforced; the report
+  catalog; the reporting area's delivery mechanism, including the
+  post-Verawood direction Phase 2 asks for.
+- **Phase 2, decided elsewhere:** permissions for custom reports (#735 — an
+  authorization fix, not a normalization decision); the survey report's own
+  design (this ADR fixes only the columns it must carry).
+- **Out:** Phase 1 (which profile fields are collected, and how the collection
+  context is recorded — a dependency here, not a subject); Phase 3 (ZIP
+  bundling); cross-course aggregation and partner-side inference.
+
+Context
+*******
+
+Each report defines its own header list independently. Course identity exists
+only in the file name (``{org}_{course}_{run}_{report}_{timestamp}.csv``), never
+in the data. The same concept is named ``Student ID``, ``User ID`` or ``id``
+across three naming styles. Row grain — what one row represents — is documented
+nowhere. Joining two reports therefore needs per-report rules re-derived by
+hand, and cross-course analysis needs file-name parsing.
+
+FCCN confirmed the base structure (Course ID, Course Run, Student/User ID, Org
+ID) for **all** existing reports, and asked whether a consolidated per-course
+"super-report" is viable.
+
+Three facts about the code shape the solution:
+
+1. **There is a single write path.** Every report, native and custom, is
+   persisted through ``upload_csv_to_report_store`` (row list),
+   ``upload_csv_file_to_report_store`` (streamed temp file) or
+   ``upload_zip_to_report_store`` (binary), in
+   ``lms/djangoapps/instructor_task/tasks_helper/utils.py``. NAU custom reports
+   reach them through ``nau_openedx_extensions/edxapp_wrapper``.
+2. **The course keys are available there; the learner key is not.** Those
+   functions receive ``course_id`` as a ``CourseKey`` (``.org``, ``.course``,
+   ``.run``). The learner identifier is per-row data.
+3. **Not every report has learner rows.** ``cohort_results`` is one row per
+   cohort with no user identity; ZIP reports have no rows at all.
+
+Two constraints bound the solution space. All work stays in
+``nau-openedx-extensions``; ``edx-platform`` is not modified unless strictly
+necessary. And the platform offers no extension point — ``instructor_task``
+exposes no ``openedx_filters`` hook of any kind.
+
+Constraints this ADR does not remove
+====================================
+
+**C1 —** ``ReportStore.links_for()`` **resolves its path without**
+``parent_dir``, so reports written to a custom directory vanish from the Data
+Download listing. *Blocks* grouping non-native reports into their own area.
+*Owner:* Phase 2, decision 8. *Resolving it delivers* a plugin-owned listing
+endpoint that reads any directory. Until it exists, ``parent_dir`` stays unused.
+
+**C2 —** ``student_profile_info`` **cannot decide its own column set.**
+``NauUserExtendedModel`` is a ``OneToOneField`` on the user, so profile answers
+are global with no record of the course or org they were collected under.
+*Blocks* freezing that report's column list. *Owner:* recording the context is
+**Phase 1**; the behaviour in its absence is decision 9 below. *Resolving it
+delivers* a profile report safe to hand to any instructor without a per-course
+review.
+
+**C3 — Verawood removes the instructor dashboard tab mechanism.** The NAU area
+today is ``FilterCertificateExportTab``, rendering a template fragment through
+``InstructorDashboardRenderStarted``. *Blocks* nothing today, but puts an expiry
+on anything built as a template tab. *Owner:* Phase 2, decision 8. *Resolving it
+delivers* a listing that survives the migration with only its presentation
+rewritten.
+
+Decision
+********
+
+Normalization means a **column contract defined per row grain**, enforced by
+wrapping the shared write path from the plugin, and documented in a catalog.
+
+**1. Two scopes, not the same list.** The *target catalog* is what NAU delivers
+(8 → 4 reports). The *column contract* covers all existing reports. They do not
+compete in cost: the course fields are injected once (point 3), so reports
+outside the catalog are covered by the contract, not by a work item.
+
+**2. Two row grains, and every report declares one.** ``learner x course run``
+carries the full base structure; ``course`` carries the course fields only, and
+the learner column is omitted rather than emitted empty. ZIP reports are outside
+the column contract and normalize on file naming only.
+
+**3. The base columns.**
+
+.. code-block::
+
+    org_id         FCT                             # course_key.org
+    course_id      course-v1:FCT+CTC101x+2020_T2   # str(course_key)
+    course_number  CTC101x                         # course_key.course
+    course_run     2020_T2                         # course_key.run
+    username       jdoe                            # learner grain only
+
+The full opaque key is emitted alongside its decomposed parts: it is the only
+globally unique join key, while the decomposed fields are what humans filter on.
+Names are ``snake_case``, lowercase, English — safe as a machine contract, since
+no report header passes through ``gettext``.
+
+**4. Injection is a wrapper installed by the plugin at startup.** The three
+report modules bind the upload functions into their own namespace at import
+time, so the plugin rebinds them there from ``AppConfig.ready()``:
+
+.. code-block:: python
+
+    # nau_openedx_extensions/reports/base_columns.py
+    #
+    # ponytail: monkeypatch, because instructor_task exposes no filter hook.
+    # Upgrade path: an openedx-filters step once one exists upstream (point 7).
+    # test_base_columns.py is what detects breakage on upgrade.
+
+    COURSE_HEADERS = ["org_id", "course_id", "course_number", "course_run"]
+    LEARNER_KEY_ALIASES = {"username", "user name", "student username"}
+
+
+    def _wrap_rows(upload):
+        @wraps(upload)
+        def wrapper(rows, csv_name, course_id, timestamp, *args, **kwargs):
+            rows = list(rows)
+            if _enabled() and rows:
+                values = _course_values(course_id)
+                rows = [COURSE_HEADERS + _normalize_header(rows[0])] + [
+                    values + list(row) for row in rows[1:]
+                ]
+            return upload(rows, csv_name, course_id, timestamp, *args, **kwargs)
+
+        return wrapper
+
+
+    def install():
+        """Called once from AppConfig.ready()."""
+        from lms.djangoapps.instructor_task.tasks_helper import (
+            enrollments, grades, misc,
+        )
+
+        for module in (enrollments, grades, misc):
+            module.upload_csv_to_report_store = _wrap_rows(
+                module.upload_csv_to_report_store
+            )
+        grades.upload_csv_file_to_report_store = _wrap_file(
+            grades.upload_csv_file_to_report_store
+        )
+
+For the streamed variant, ``_wrap_file`` wraps the file in a view that prepends
+a CSV-escaped constant prefix to every line, so the report is never re-parsed.
+That is a correctness and speed argument, not a memory one: ``store()`` already
+does ``buff.read()`` on the whole file.
+
+``upload_zip_to_report_store`` is deliberately left unwrapped.
+
+NAU custom reports need no patching — they already import through
+``edxapp_wrapper``, their legitimate injection point.
+
+**5. The wrapper also renames the learner key**, via the alias map above. This
+is what makes the fourth field achievable without editing any report: the value
+is already present in every learner-grain report, only its column name differs.
+Consequently **the whole normalization is one module plus one test**, and
+per-report work for the target catalog is close to zero.
+
+The ceiling is stated openly: a monkeypatch breaks silently if upstream renames
+those modules, changes the import style, or adds a fourth report module.
+
+**6. Base columns are prepended, and the change is breaking.** Consumers reading
+by column position break at the cutover, announced with version 1 of the
+catalog. The behaviour ships behind a plugin setting, defaulting to off — a
+rollback that needs no code change.
+
+**7. Upstream is a follow-up, not the delivery vehicle.** What NAU proposes
+upstream is an **extension point** — a filter on report upload, in an area that
+has none — so the monkeypatch can be retired. Delivery does not wait for it.
+
+**8. The reporting area is delivered as an API first, a tab second.** The
+current tab launches tasks but does not list reports, so a listing must be built
+regardless. It is implemented as a REST endpoint in the plugin honouring
+``parent_dir`` (resolving C1); the existing tab consumes it today, an MFE slot
+consumes it after Verawood (resolving C3). The architectural recommendation FCCN
+asked for is then a description of what was built, not a separate study.
+
+**9. Profile data with no recorded collection context is omitted**, consistent
+with deny-by-default. Recording the context is Phase 1; this is only the
+report's behaviour in its absence.
+
+**10. One conformance test asserts the leading columns of every report.** Under
+point 4 it is also the only detector of a monkeypatch upstream has invalidated,
+so it is mandatory. It runs in the plugin's suite against the pinned platform
+version.
+
+**11. The certificate issue date is delivered by the join, not a new column.**
+Issue #32's value already exists in ``export_course_certificates`` as
+``certificate created date``. Once both reports carry the base structure they
+join on ``course_id`` plus the learner key, and the requirement is met with no
+code. If the column must be physically inside ``grade_report``, the route is a
+small upstream contribution to ``_user_certificate_info`` — not a plugin patch,
+since the grade report's internal header and row builders change shape between
+releases while the upload signature does not.
+
+**12. The report catalog is a deliverable:** per report, name, row grain, base
+fields, full column list, and the variables still requiring partner-side
+inference (for example "course status", which has no single native field).
+
+Consequences
+************
+
+- Every CSV grows by four to five columns.
+- Consumers reading by column position break at the cutover; those mapping by
+  header name are unaffected.
+- NAU carries a startup monkeypatch against three upstream namespaces — one
+  plugin file, gated by a setting, covered by the conformance test. Every Open
+  edX upgrade must run that test before release: an upstream rename disables
+  the injection silently, and reports keep generating without the base
+  structure.
+- Cross-report joins stop needing file-name parsing; cross-course aggregation
+  needs no further platform work.
+- Reports must not move into the NAU area before the listing endpoint exists.
+- Deny-by-default has a visible effect: learners who registered before Phase 1
+  records a collection context lose their profile columns unless a backfill is
+  agreed. FCCN must hear this before discovering it in a report.
+- ZIP *bundling* (Phase 3) is not the same subject as reports that already are
+  ZIP archives; the two should not be conflated in planning.
+
+Rejected Alternatives
+*********************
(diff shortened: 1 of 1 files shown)


efortish:
Hello @cabralpedrofccn 

I'd like to get a review on this ADR, it defines a 1st starting point to execute the phase 2 of ARTE project; I left a few open questions that I'd like to discuss with NAU team before proceding with this 2nd phase.

@mariajgrimaldi I'd like to know what you think about it.

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+
+**Nobody has written down what a row is.** Most reports have one row per
+learner per course run. But ``cohort_results`` has one row per cohort, and some
+reports are ZIP archives with no rows at all.
+
+The result is that joining two reports needs rules invented per report, and
+cross-course analysis needs file-name parsing. FCCN has asked for a mandatory
+base structure — Course ID, Course Run, Student/User ID, Org ID — in every
+report, and has also asked whether a consolidated per-course "super-report" is
+viable.
```
mariajgrimaldi: I believe the super-report simplification would be out of scope as discussed in the Basecamp issue.

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+
+1. Reports declare their row grain
+==================================
+
+Two grains exist:
+
+- **learner grain** — one row per learner per course run. Carries all base
+  columns.
+- **course grain** — one row about the course run itself. Carries the course
+  columns; the learner column is omitted, not left empty.
```
mariajgrimaldi: I think an example would be useful here.

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+
+ZIP reports have no rows. They are outside the column contract and normalize on
+file naming only.
```
mariajgrimaldi: What does this mean?

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+
+Consumers that read by column position will break. The change is announced with
+a dated cutover and version 1 of the report catalog. It ships behind a plugin
+setting that defaults to off, so rollback needs no code change.
+
+5. Profile data without a recorded collection context is omitted
+================================================================
+
+This is deny-by-default. Recording that context is Phase 1 work; this ADR only
+decides what the report does when it is missing.
```
mariajgrimaldi: This might need a bit more context? What I understand is: profile data is not included, the only personal identifiable information included will be username?

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+
+6. The certificate issue date is delivered by a join, not a new column
+======================================================================
+
+Issue #32 asks for the certificate date in ``grade_report``. That value already
```
mariajgrimaldi: I think we could link the issue here.

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+exists in ``export_course_certificates``. Once both reports carry the base
```
mariajgrimaldi: What's the other report we're considering here? Not sure I'm understanding this decision.

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+- ``student_profile_info`` builds its header from
+  ``student_profile_download_fields``, a site configuration value that
+  *replaces* the default field list. A deployment configured without
+  ``username`` would silently stop satisfying the contract, so the conformance
+  test must cover that case.
+
+The conformance test
+====================
+
+One test runs each report generator and asserts its leading columns. It is
```
mariajgrimaldi: What test are we referring to here?

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+Consequences
+************
+
+- Every CSV grows by four to five columns.
+- Consumers reading by column position break at the cutover. Consumers mapping
+  by header name are unaffected.
+- Every Open edX upgrade must run the conformance test before release.
+- Cross-report joins no longer need file-name parsing, and cross-course
+  aggregation needs no further platform work.
+- Learners who registered before Phase 1 records a collection context will lose
```
mariajgrimaldi: I'm not sure why they would lose these records, can we expand on this please?

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+
+Dependencies
+************
+
+These are not solved by the column contract.
+
+**The platform's report listing ignores** ``parent_dir``.
+``ReportStore.links_for()`` resolves its path without it, so a report written to
+a custom directory disappears from the Data Download listing. Until the plugin
+has its own listing endpoint, ``parent_dir`` stays unused.
```
mariajgrimaldi: What's the current consequence of this?

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+
+**Verawood removes the instructor dashboard tab mechanism.** The NAU area today
+is ``FilterCertificateExportTab``, a template fragment rendered through
+``InstructorDashboardRenderStarted``. Anything built as a template tab has an
+expiry date. The listing should therefore be built as a REST endpoint that the
+current tab consumes now and an MFE slot consumes later. That endpoint also
+resolves the ``parent_dir`` limitation above, and it is the architectural
+recommendation FCCN asked for as a Phase 2 deliverable.
```
mariajgrimaldi: Yes, agreed. Although I don't think we should build the listing right away, I think it'd be too much work for the immediate gain. I think a migration plan would be enough for now.

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+
+**Profile data has no collection context.** ``NauUserExtendedModel`` is a
```
mariajgrimaldi: What is collection context?

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+``OneToOneField`` on the user, so profile answers are global. Until Phase 1
+records where each answer was collected, the column set of
+``student_profile_info`` cannot be frozen.
```
mariajgrimaldi: I think I need further examples to understand this better.

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+or drop rows. Grade reports have a per-course variable column count, so the
+consolidated schema would differ per course, which is the opposite of
+normalization. It would also cost the heaviest report plus the join on every
+request.
+
+Open Questions
+**************
+
+**Where do the reports live?** The scope email says the survey report goes in
+"the same reporting area instructors already use". The issue's acceptance
```
mariajgrimaldi: They refer to the instructor dashboard tab I believe. Would this be a problem?

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+criteria call for a dedicated NAU area with a different name. If reports stay in
+the standard listing, ``parent_dir`` is unused and the listing work shrinks to
+the post-Verawood recommendation alone. *This has the largest effect on effort.*
+
+**Does "Course ID" mean the full opaque key or the course number?** We emit
```
mariajgrimaldi: Course ID is the course key string.

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+both; a confirmation would let us drop one.
+
+**Is** ``username`` **acceptable under GDPR?** Technically it is settled. The
```
mariajgrimaldi: I think under GDPR username's are PII (personal identifiable information). If we all settle on this, then we should try a different ID, but I also think it should be an explicit decision part of this ADR.

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+question is whether joins should run on the course-specific anonymous id
+instead, with identity confined to ``student_profile_info``.
+
+**Does ARTE map columns by header name or by position?** This decides whether
+the change is a release note or a dated cutover.
```
mariajgrimaldi: This is a new feature affecting all reports so the bump will be major either way.

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+
+**Is a three-field base structure accepted for course-grain reports?**
+
+**Is a backfill of profile collection context required for existing learners?**
```
mariajgrimaldi: Not sure what this means.

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+
+**Is the certificate date acceptable as a join, or must it be a real column?**
```
mariajgrimaldi: A join with?

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,326 @@
...
+
+**Is "join keys instead of a super-report" accepted?**
```
mariajgrimaldi: It is under the current scope.

mariajgrimaldi (changes requested):

docs/decisions/0001-report-normalization.rst
efortish: @cabralpedrofccn 
Could you please bring us that answer?

docs/decisions/0001-report-normalization.rst
efortish: It is export_course_certificates, which already carries the value as certificate created date.

docs/decisions/0001-report-normalization.rst
efortish: That is a concern I already shared with Paola to talk with the NAU team.

docs/decisions/0001-report-normalization.rst
efortish: Agreed, and you are right that it deserves its own decision.

docs/decisions/0001-report-normalization.rst
efortish: True, removed that framing. I kept the question about how ARTE maps columns, since it changes what migration support they need,

docs/decisions/0001-report-normalization.rst
efortish: With export_course_certificates.

docs/decisions/0001-report-normalization.rst
efortish: Reworded. The archives are Phase 2. The point is that they do not need columns of their own: export_course_certificates already has one row per certificate including certificate_download_pdf_link, so once that CSV carries the base structure every PDF in the archive is reachable from a normalized row. That is separate from ZIP bundling, which moved to Phase 3.

docs/decisions/0001-report-normalization.rst
efortish: Since we are not using parent_dir at all, every report is listed the same way and this never comes up. I removed the dependency. If we ever move reports to their own directory it belongs in the migration plan, not here.

docs/decisions/0001-report-normalization.rst
efortish: Added more context. The wrapper rebinds functions inside three edx-platform modules we do not control. If an upgrade renames those modules or changes how they import, the rebinding silently does nothing: reports keep generating, just without the base columns, and nobody notices until the partner reports missing data. The test runs each report against a fixture course and asserts the base columns are present, so that breakage shows up during the upgrade instead.

docs/decisions/0001-report-normalization.rst
efortish: Removed. Decision 1 already settles it, so asking it again at the end was just confusing. Course grain reports carry org_id, course_id and course_run and no learner key, and if FCCN disagrees that is a change to decision 1.

efortish:
@mariajgrimaldi 

I went back to Phase 1 (#[954](https://github.com/fccn/nau-technical/issues/954)), and I think there is a real contradiction between the two issues, so I reworked Decision 6 rather than making an assumption.

Phase 1 puts the NAU fields on `NAUUserExtendedModel`, which is associated with the user account, and collects them during registration. The gate before course access only exists to backfill learners who registered before these fields were introduced. This means the data is account-level by design.

However, #[955](https://github.com/fccn/nau-technical/issues/955) asks that data provided for course 1 / org XPTO must not appear in a report for course 2 / org XYZ. That only makes sense if the fields are collected at the course or org level, which is not how Phase 1 is currently designed. Nothing currently records which course or org an answer came from, so we cannot filter by that context today, even if we wanted to.

The two options lead to very different implementations:

* **Account-level:** `student_profile_info` reports the fields for any course the learner is enrolled in. This requires no additional work and matches what Phase 1 is already building.
* **Per-course or per-org:** We would need to store the context associated with each answer, decide how to handle learners who have already provided answers, and add filtering to the report. This would require a change to the Phase 1 data model rather than something Phase 2 could absorb.

I left this as an open question in the ADR because it blocks us from freezing the column set for `student_profile_info`. It would be worth asking Pedro directly which of these two behaviors he intended, since he raised requirements that point in both directions.


@cabralpedrofccn  could you please take a look of this?

mariajgrimaldi:
@efortish: here are my two cents

Although the data is associated with the account, the way we usually connect students with courses is through enrollments. As I see it, if we only include students enrolled in the course the report belongs to, we’d comply with the requirement: _Course 1 / org XPTO must not appear in a report for Course 2 / org XYZ_.

For example, if Student A is only enrolled in Course 1, their data would only appear in the report for Course 1. If Student B is only enrolled in Course 2, their data would only appear in the report for Course 2. If Student C is enrolled in both courses, their data would appear in both reports, which would also be correct regardless of which course they were taking when they filled in the data.

Basically, the data remains associated with the account, but enrollment determines which course reports the student is included in.

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,379 @@
...
+
+Scope: Phase 2 only, and within it only what report normalization means and how
+it is built. Phase 1 (profile field collection) and Phase 3 (ZIP bundling) are
+dependencies, not subjects.
+
+Context
+*******
+
+Every course report defines its own columns independently. This causes three
+problems.
```
mariajgrimaldi:
```suggestion
Every course report defines its own columns independently. This causes three
problems:
```

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,379 @@
...
+point. The plugin therefore installs a wrapper at startup. See *Implementation*
+below.
+
+5. Base columns go first
+========================
+
+Consumers that read by column position will break. This is a new feature
+affecting every report, so the release is a major version bump regardless, and
+the change is announced with version 1 of the report catalog. It ships behind a
+plugin setting that defaults to off, so rollback needs no code change.
```
mariajgrimaldi: What are we gaining by prepending instead of appending?

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,379 @@
...
+If a learner column is found, the report is learner grain, and the wrapper
+resolves each username to ``anonymous_id_for_user(user, course_id)``. The
+mapping is built once per report with a single query over the course's enrolled
+users, not once per row.
+
+If no learner column is found, the report is course grain and only the course
+columns are added.
+
+Because of this, no report needs to be edited to satisfy the contract. The whole
+normalization is one module and one test.
```
mariajgrimaldi: If the current report uses username, what's the tradeoff of adding more calculations to get the anonymous ID?

docs/decisions/0001-report-normalization.rst
```diff
@@ -0,0 +1,379 @@
...
+
+Two adjustments are still required
+==================================
+
+- ``export_course_certificates`` already emits its own ``course_id`` column.
+  Once the wrapper adds the base structure, the file would carry that header
+  twice. The report is NAU-owned, so its own column is dropped. This must land
+  in the same change as the wrapper.
```
mariajgrimaldi: I don't think we should drop but avoid adding it

docs/decisions/0001-report-normalization.rst
mariajgrimaldi: Do we have any updates?

docs/decisions/0001-report-normalization.rst
mariajgrimaldi: (non blocking)

docs/decisions/0001-report-normalization.rst
mariajgrimaldi: (non blocking)

docs/decisions/0001-report-normalization.rst
mariajgrimaldi: (non blocking)

docs/decisions/0001-report-normalization.rst
efortish: This one is gone. It was a consequence of the collection-context model, where a learner only got profile columns in a report if they had filled the fields while in that course's context, so anyone who registered before Phase 1 started recording a context would have had no context to match and would have dropped out of the report.

That model is no longer in the ADR. The fields are account data and reports are scoped by enrollment, so a learner who has filled them appears with them in every course they are enrolled in, whenever they filled them. Nobody loses records.

docs/decisions/0001-report-normalization.rst
efortish: That term is gone from the ADR. It came from an earlier draft that assumed the profile fields were captured per course, so each value carried the course or org it was collected under, and reports would filter on it.

Phase 1 does not collect them that way, so the concept had no basis. Decision 6 now says the fields are account data and scoping is done by enrollment.

docs/decisions/0001-report-normalization.rst
efortish: Fair, and rather than add examples I removed the thing that needed them. The passage described the collection-context model, which is not what Phase 1 builds. Decision 6 is now a page shorter and turns on one rule: a course report covers the learners enrolled in that course.

docs/decisions/0001-report-normalization.rst
efortish: Applied.

docs/decisions/0001-report-normalization.rst
efortish: Good question, the ADR did not say. Added to decision 5.

The gain is position. Put first, the base block sits at the same index in every report, so a consumer can read it without knowing which report it is holding, which is most of what "normalized" is buying here.

Appended, they would land at a different index per report, and `grade_report` already has a column count that varies per course, so there is no stable trailing position to append to either. Appending would cost the same break and give less.

docs/decisions/0001-report-normalization.rst
efortish: Two separate costs, and both are already in the text, just far apart.

Compute is one query. The wrapper builds the username to anonymous ID map once per report, over the course's enrolled users, not once per row, so it does not scale with report size.

The real trade is in decision 3: `username` is the cheapest option because all four reports already carry it, but it is PII travelling into the partner's systems, and it is mutable, since learners can be renamed, which makes it a bad join key regardless of GDPR.

docs/decisions/0001-report-normalization.rst
efortish: I went with the drop, and added the reasoning to that bullet.

Skipping keeps `export_course_certificates` untouched, which is the appeal, but it leaves that `course_id` wherever the report puts it, in the middle of the row. Decision 5 promises the base block in the same place in every report, and this would be the one report where that is false, which is worse than the thing it avoids.

On the cost: yes, it is only this report, and it is one line in a file NAU owns. Nothing upstream is touched.

docs/decisions/0001-report-normalization.rst
efortish: Splitting this in two, because half of it is settled and half is not.

Settled: the join key this contract adds is not the username. Decision 3 uses `anonymous_id_for_user(user, course_id)`, so the column the partner joins on carries no identity. The username is read to resolve the key and is not published as one, so normalization itself adds no PII to any report.

Not settled, and this is the part that still needs FCCN: reports that already emit `Username` or `Email` keep them, because this contract adds columns and does not remove any. `student_profile_info` and `export_course_certificates` both ship identity today, before and after this change. Whether they should is a per-report decision that predates this ADR, so I kept it out of the contract rather than deciding it here.

@cabralpedrofccn, that second half is the one we need from you: are the existing `Username` and `Email` columns acceptable in the reports that go to the partner? Leaving this thread open until we hear back.

docs/decisions/0001-report-normalization.rst
efortish: No update from NAU yet, so still open.

To be clear about what it does and does not block: the ADR takes the conservative position already, decision at the end of *Constraints* says we do not build a separate reporting area or listing endpoint now, precisely because `FilterCertificateExportTab` is a template fragment through `InstructorDashboardRenderStarted` and Verawood removes that mechanism. Reports stay in the standard Data Download listing.

So nothing in this ADR depends on the answer. What needs it is the decision to move reports into a NAU-owned area later, which is not Phase 2 work. I would rather merge this and track the tab question separately than hold the contract on it.

@paolacruzedunext, any word from NAU on this one?

efortish:
Pushed 8a9d8d4. @mariajgrimaldi, all your threads are answered, eight resolved and two left open on purpose.

**The scoping question is closed**, using what you proposed on 25 August. The fields stay account data and each report covers the learners enrolled in that course, which satisfies what the issue asked for without any per-course filtering. Decision 6 says that now and the open question is gone.

I added the consequence in plain words, because FCCN should hear it from us rather than find it: a learner enrolled at two organizations shows the same profile values in both reports. There is one value per learner, and scoping by enrollment does not change that. It is the trade Phase 1 made when it chose to collect these fields once per learner.

**The collection-context threads are obsolete.** Three of them (why learners would lose records, what collection context is, needing examples) all described a model where each value carried the course it was captured under. Phase 1 does not build that, so the passage is gone rather than clarified.

**The two design questions are now answered in the text**, not just in a reply. Decision 5 says why the base columns go first: appended, they would sit at a different index per report, and `grade_report`'s column count already varies per course, so there is no stable trailing position to append to. And on `export_course_certificates`, I kept the drop over having the wrapper skip, since skipping leaves that `course_id` mid-row and makes this the one report where the base block is not where the contract says it is. The bullet explains that.

**Two threads stay open, both waiting on NAU, not on us.** Whether the existing `Username` and `Email` columns are acceptable in the reports the partner receives, which is a per-report question this contract deliberately does not decide, and the instructor dashboard tab. Neither blocks the ADR: the contract adds no PII of its own, and it already takes the conservative line on the tab by not building a NAU reporting area.

Also fixed the title underline, which was three characters short.

PR #157, opened 2026-08-20
fix: recover question text for div-wrapped MCQ/checkbox problems

rguerra-fccn:
## Summary

Follow-up to fccn/nau-technical#948. After the "Resposta"/"Resposta Correta" fix (#154) was deployed to production, Carolina reported that the "Pergunta" (question) column in the "Problem Responses" report always shows a generic "Question N" / "Questão N" placeholder instead of the actual question text.

- fccn/nau-technical#948
- https://github.com/fccn/nau-openedx-extensions/pull/154

## Root cause

`LoncapaProblem.find_question_label` (edx-platform core, `xmodule/capa/capa_problem.py`) looks for the question prompt (a `<p>`/`<label>` element) as the sibling immediately **preceding the response tag itself**, e.g.:

```xml
<p>What is 2+2?</p>
<multiplechoiceresponse>
    <choicegroup id="...">...</choicegroup>
</multiplechoiceresponse>
```

But Studio's rich-text/per-choice-feedback editor (the same editor implicated in #948) instead nests the prompt **inside** the response tag, as a sibling of the choice group:

```xml
<multiplechoiceresponse>
    <p>What is 2+2?</p>
    <choicegroup id="...">...</choicegroup>
</multiplechoiceresponse>
```

so the original lookup walks up one level too many and never finds it, falling through to the "Question N" default.

A second, independent bug affects even the "expected" (sibling-of-response-tag) layout: the prompt's text is read via the element's direct `.text` node only, so a prompt authored with any inline rich-text formatting (e.g. `<p><strong>What is 2+2?</strong></p>`) also resolves to empty/`None`.

A third case, found while verifying coverage across all question types in the production course: "matching"/dropdown problems (a single `optionresponse` with several `optioninput` sub-answers) author each sub-answer's own prompt as a `<div>` sibling immediately preceding it, not a `<p>`/`<label>` — the original recovery logic didn't recognise `<div>` as a valid prompt wrapper either, so these also fell back to "Question N" for every sub-answer.

## Fix

Patches `find_question_label` (following the same monkeypatch approach already used in this module for `find_answer_text`/`find_correct_answer_text`) to:
1. Also look for the prompt as a sibling of the answer element itself, in addition to the original sibling-of-response-tag lookup.
2. Extract text recursively (via `itertext()`) instead of relying on the direct `.text` node.
3. Recognise `<div>` as a valid prompt-wrapper element, in addition to `<p>`/`<label>`, to also recover per-sub-answer prompts in `optionresponse`/matching problems.

All fall back to the original result if no real question text can be recovered, so behaviour for already-working problems is unchanged.

## Testing

- Added unit tests (`TestFindQuestionLabelFactory`, 7 tests) covering: passthrough when original result is real text, recovery for the nested-in-response-tag layout (with rich-text markup), recovery for the legacy layout with rich-text markup, no-match fallback, no-recovery-possible fallback, skipping `<description>` elements, and recovery from `<div>`-wrapped prompts in a multi-sub-answer `optionresponse`. All 17 tests in the file pass.
- `pycodestyle`, `pylint`, and `isort` all clean on the changed files.
- Verified end-to-end against the real production OLX (the `CiênciaVitae` problem from the original bug report, in the `course-v1:FCT+SERVDIG+2025_T3` test course in the local dev environment): confirmed the bug reproduces without the fix ("Questão 1"), and is resolved with the fix ("O que é o CiênciaVitae?"), calling `generate_report_data()` directly (the actual production code path used by the CSV report generator).
- **Coverage across all question types**: checked every problem in the production course tarball (36 files: 25 `multiplechoiceresponse`, 10 `choiceresponse`/checkbox, 1 `optionresponse`) — 39 answer_ids total. With this fix, **all 39 now correctly recover their real question text**, with **zero regressions**.
- **Regression check on the already-deployed #948 fix** (`extract_choices`/`find_correct_answer_text`, unrelated to this PR but re-verified while investigating question-type coverage): confirmed via synthetic OLX + real `LoncapaProblem` instantiation that `numericalresponse`, `stringresponse`, `formularesponse`, `optionresponse`, and plain/legacy (non-rich-text) MCQ problems all behave identically to the unpatched upstream — the patch's own guard clauses already exclude these response types from its recovery logic.

TODO(nau-technical#948): This is a stopgap fix for an upstream Open edX bug. Once an upstream fix lands in edx-platform/xblocks-contrib, remove these monkeypatches.

### DIFF
--- a/nau_openedx_extensions/apps.py
+++ b/nau_openedx_extensions/apps.py
@@ -67,8 +67,9 @@ class NauOpenEdxConfig(AppConfig):
             get_course_email_context_factory(prev_email_context_func)  # pylint: disable=protected-access
 
         # Fix the "Problem Responses" report showing "Answer Text Missing" / an empty
-        # correct answer for multiple-choice/checkbox problems whose choices were authored
-        # with Studio's rich-text / per-choice feedback editor (see nau-technical#948).
+        # correct answer, and "Question N" instead of the actual question text, for
+        # multiple-choice/checkbox problems whose choices were authored with Studio's
+        # rich-text / per-choice feedback editor (see nau-technical#948).
         # TODO(nau-technical#948): remove once fixed upstream in edx-platform/xblocks-contrib.
         from xmodule.capa.capa_problem import \
             LoncapaProblem  # pylint: disable=import-error,import-outside-toplevel # noqa
@@ -77,6 +78,7 @@ class NauOpenEdxConfig(AppConfig):
         from nau_openedx_extensions.xblocks.capa_problem_responses import (  # pylint: disable=import-outside-toplevel # noqa
             get_extract_choices_factory,
             get_find_correct_answer_text_factory,
+            get_find_question_label_factory,
         )
 
         prev_extract_choices_func = ChoiceGroup.extract_choices  # pylint: disable=protected-access
@@ -87,3 +89,7 @@ class NauOpenEdxConfig(AppConfig):
         prev_find_correct_answer_text_func = LoncapaProblem.find_correct_answer_text  # pylint: disable=protected-access
         LoncapaProblem.find_correct_answer_text = \
             get_find_correct_answer_text_factory(prev_find_correct_answer_text_func)  # pylint: disable=protected-access
+
+        prev_find_question_label_func = LoncapaProblem.find_question_label  # pylint: disable=protected-access
+        LoncapaProblem.find_question_label = \
+            get_find_question_label_factory(prev_find_question_label_func)  # pylint: disable=protected-access
--- a/nau_openedx_extensions/tests/test_capa_problem_responses.py
+++ b/nau_openedx_extensions/tests/test_capa_problem_responses.py
@@ -400,3 +400,78 @@ class TestFindQuestionLabelFactory(unittest.TestCase):
 
         self.assertEqual(wrapper(self._make_lcp(tree), 'x_2_1'), 'First item prompt')
         self.assertEqual(wrapper(self._make_lcp(tree), 'x_2_2'), 'Second item prompt')
+
+    def test_ignores_non_prompt_div_for_non_optioninput_answers(self):
+        """
+        `<div>` recovery is intentionally scoped to `<optioninput>` answers
+        only (see `_find_label_element_text`'s `allow_div` docstring): for
+        other response types (e.g. MCQ/checkbox), a `<div>` immediately
+        preceding the answer element is common for unrelated purposes (an
+        image wrapper, a shared/general instructions block, etc.), so it
+        must NOT be picked up as the question prompt -- doing so would
+        silently substitute incorrect text into the report instead of the
+        (at least honest) "Question N" placeholder.
+        """
+        tree = _parse(
+            '<problem><multiplechoiceresponse>'
+            '<div>Some unrelated image caption, not a question prompt.</div>'
+            '<choicegroup id="x_2_1">'
+            '<choice correct="true"><div>Answer</div></choice>'
+            '</choicegroup>'
+            '</multiplechoiceresponse></problem>'
+        )
+        original_func = Mock(return_value='Question 1')
+        wrapper = get_find_question_label_factory(original_func)
+
+        result = wrapper(self._make_lcp(tree), 'x_2_1')
+
+        self.assertEqual(result, 'Question 1')
+
+    def test_recovers_question_even_with_malformed_answer_id(self):
+        """
+        The generic-default detection must not depend on `answer_id`
+        matching the platform's usual `..._<n>_<m>` format: it's derived
+        from a regex over the translated template, not from parsing
+        `answer_id` itself, so recovery still works even for an unexpected
+        answer_id shape.
+        """
+        tree = _parse(
+            '<problem><multiplechoiceresponse>'
+            '<p>What is the real question?</p>'
+            '<choicegroup id="not-the-usual-id-format">'
+            '<choice correct="true"><div>Answer</div></choice>'
+            '</choicegroup>'
+            '</multiplechoiceresponse></problem>'
+        )
+        original_func = Mock(return_value='Question 1')
+        wrapper = get_find_question_label_factory(original_func)
+
+        result = wrapper(self._make_lcp(tree), 'not-the-usual-id-format')
+
+        self.assertEqual(result, 'What is the real question?')
+
+    def test_falls_back_gracefully_when_i18n_gettext_raises(self):
+        """
+        If `capa_system.i18n.gettext` is missing/raises an AttributeError or
+        TypeError for some reason (e.g. the i18n service itself being
+        unavailable/misconfigured), the wrapper must not propagate the
+        exception: it should log a warning and fall back to returning the
+        original (non-empty) result unmodified, exactly as if recovery had
+        been skipped.
+        """
+        tree = _parse(
+            '<problem><multiplechoiceresponse>'
+            '<p>Real question text</p>'
+            '<choicegroup id="x_2_1">'
+            '<choice correct="true"><div>Answer</div></choice>'
+            '</choicegroup>'
+            '</multiplechoiceresponse></problem>'
+        )
+        original_func = Mock(return_value='Question 1')
+        wrapper = get_find_question_label_factory(original_func)
+        lcp = self._make_lcp(tree)
+        lcp.capa_system.i18n.gettext = Mock(side_effect=AttributeError('i18n unavailable'))
+
+        result = wrapper(lcp, 'x_2_1')
+
+        self.assertEqual(result, 'Question 1')
--- a/nau_openedx_extensions/tests/test_capa_problem_responses.py
+++ b/nau_openedx_extensions/tests/test_capa_problem_responses.py
@@ -375,3 +375,28 @@ class TestFindQuestionLabelFactory(unittest.TestCase):
         result = wrapper(self._make_lcp(tree), 'x_2_1')
 
         self.assertEqual(result, 'Real question text')
+
+    def test_recovers_question_from_div_wrapped_prompt(self):
+        """
+        Reproduces a real production "matching" problem: a single
+        <optionresponse> containing several <optioninput> sub-answers, each
+        with its own <div> prompt immediately preceding it (not <p>/<label>).
+        The original implementation only recognises <p>/<label>, so it falls
+        back to the generic default for every sub-answer; the wrapper must
+        also recognise <div> and recover each sub-answer's own prompt.
+        """
+        tree = _parse(
+            '<problem><optionresponse>'
+            '<div>Overall instructions, not a specific prompt.</div>'
+            '<div>First item prompt</div>'
+            '<optioninput id="x_2_1"><option correct="true">A</option></optioninput>'
+            '<br/>'
+            '<div>Second item prompt</div>'
+            '<optioninput id="x_2_2"><option correct="true">B</option></optioninput>'
+            '</optionresponse></problem>'
+        )
+        original_func = Mock(return_value='Question 1')
+        wrapper = get_find_question_label_factory(original_func)
+
+        self.assertEqual(wrapper(self._make_lcp(tree), 'x_2_1'), 'First item prompt')
+        self.assertEqual(wrapper(self._make_lcp(tree), 'x_2_2'), 'Second item prompt')
--- a/nau_openedx_extensions/tests/test_capa_problem_responses.py
+++ b/nau_openedx_extensions/tests/test_capa_problem_responses.py
@@ -401,6 +401,56 @@ class TestFindQuestionLabelFactory(unittest.TestCase):
         self.assertEqual(wrapper(self._make_lcp(tree), 'x_2_1'), 'First item prompt')
         self.assertEqual(wrapper(self._make_lcp(tree), 'x_2_2'), 'Second item prompt')
 
+    def test_recovers_question_from_div_wrapped_prompt_with_trailing_br(self):
+        """
+        Same "matching" problem shape as `test_recovers_question_from_div_wrapped_prompt`,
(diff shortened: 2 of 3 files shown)


ManuelStarDo (changes requested):
## Review

**[HIGH]** `nau_openedx_extensions/xblocks/capa_problem_responses.py:203` (`LABEL_ELEMS = ('p', 'label', 'div')`) — treating any preceding `<div>` as a question prompt is a much broader match than `<p>`/`<label>` and risks false positives.
Risk: Unlike `<p>`/`<label>`, `<div>` is a generic, extremely common container tag used throughout capa OLX for many unrelated purposes — layout wrappers, images, embedded HTML, feedback/explanation blocks that aren't tagged `<description>`, etc. Because `_find_label_element_text` is invoked for *every* answer_id whose original result looks like a default (not just `optionresponse`/matching problems), any MCQ/checkbox/other problem that happens to have a non-prompt `<div>` immediately preceding the response tag or the answer element (e.g. an image caption, a spacer, a partially-migrated HTML block) will now have that div's text silently substituted in as the "question," producing incorrect (not just missing) data in the report — arguably worse than the current "Question N" placeholder, since it looks plausible but is wrong. The PR's own verification (39 answer_ids across 36 files) is reassuring but is only a sample of the current production course, not a guarantee for other courses/future content — and there's no guard limiting the `<div>` match to `optionresponse`-specific structures (e.g. checking `xml_element.tag == 'optioninput'` or that the parent is `optionresponse`) even though that's the only scenario described/tested for the `<div>` case.
Fix: Scope the `<div>` match narrower — e.g., only allow `<div>` as a valid prompt wrapper when the target element (or its parent) is specifically an `optioninput`/`optionresponse`, mirroring how the "legacy" vs "nested" branches are already dispatched based on structure. Alternatively, add a length/heuristic sanity check (e.g., skip candidate divs containing nested ``/`<optioninput>`/interactive elements) or require the div to be the *sole* preceding sibling content specific to that answer (not shared/overall instructions). At minimum, add a test asserting a non-prompt `<div>` (e.g. an image wrapper or a shared instructions block, as already partially modeled by "Overall instructions, not a specific prompt" in the new test tree) is **not** picked up when it precedes an MCQ/checkbox answer directly (the new test only checks the positive `optionresponse` case, not a negative case for other problem types).


**[HIGH]** `nau_openedx_extensions/xblocks/capa_problem_responses.py:254` — sentinel-based default-detection (`result != default_label`) is fragile: if `answer_id.split('_')[-2]` doesn't parse as expected, `default_label` silently becomes `None`, and recovery is skipped entirely with no logging. Still present, unchanged.
Fix: (as before) log the `except` branch, and/or validate `result` against a `Question \d+` pattern rather than exact reconstructed-string equality.

**[LOW / SUGGESTION]** No test covers the malformed-`answer_id` sentinel-mismatch path. Still an open gap.

**[LOW / SUGGESTION]** The `gettext_fn` babel-extraction-avoidance comment is unverified against this repo's actual i18n config. Still unverified, still low risk.

rguerra-fccn:
Thanks for the thorough review, @ManuelStarDo! All four points have been addressed in `adf6501` (pushed to this branch):

**[HIGH] `<div>` scoping too broad**
Scoped the `<div>` match so it's only ever considered a valid prompt wrapper when the target element is specifically an `<optioninput>` — the only verified scenario for this. Added a new `allow_div` parameter to `_find_label_element_text` (default `False`); the call site now passes `allow_div=xml_element.tag == 'optioninput'`, so MCQ/checkbox/other problem types never fall into the `<div>` branch, regardless of what unrelated `<div>`s (images, layout wrappers, feedback blocks) happen to precede them. Also added the negative test you asked for: `test_ignores_non_prompt_div_for_non_optioninput_answers`, confirming a `<div>` preceding an MCQ/checkbox answer is **not** picked up as the prompt.

**[HIGH] Fragile sentinel-based default-detection**
Replaced the `answer_id.split('_')[-2]` exact-string reconstruction with a regex-based check: the translated template `gettext_fn("Question {}")` is split into its literal prefix/suffix around the `{}` placeholder, and `result` is matched against `re.fullmatch(re.escape(prefix) + r'-?\d+' + re.escape(suffix), result)`. This detects *any* generic "Question N" default regardless of whether `answer_id` happens to parse into the expected `..._<n>_<m>` shape. The whole check is wrapped in `try/except (AttributeError, TypeError)` with `log.warning(..., exc_info=True)` on failure (covering a missing/broken `capa_system.i18n` service or a non-callable/non-string `gettext_fn` return), falling back to *not* attempting recovery rather than failing silently or raising.

**[LOW] No test for the malformed-`answer_id` path**
Added `test_recovers_question_even_with_malformed_answer_id`, confirming recovery still works correctly when `answer_id` doesn't match the usual format (this is actually now a non-issue by construction, since the new detection no longer depends on parsing `answer_id` at all — but the test locks that in). Also added `test_falls_back_gracefully_when_i18n_gettext_raises` covering the exception-handling fallback path itself.

**[LOW] Unverified babel-extraction claim**
Empirically verified rather than just asserted: temporarily renamed `gettext_fn` back to `_` and ran the repo's actual `pybabel extract -F nau_openedx_extensions/locale/babel.cfg ...` command — confirmed it **does** add a new `msgid "Question {}"` entry with `_`, and does **not** with `gettext_fn`. Updated the code comment to reflect this was confirmed via that exact invocation.

Also fixed an unrelated CI lint failure introduced by the first round of these changes (pylint `W0718 broad-exception-caught` on the new `except` block) by narrowing it to `(AttributeError, TypeError)` as noted above.

Full regression check: re-ran the 36-file/39-`answer_id` production OLX sample after all these changes — still 39/39 recovered, 0 regressed, 0 still-broken. All 20 unit tests pass, and `pylint`/`pycodestyle`/`isort` are clean against the CI's exact full-package invocation. CI is green on this PR again.

ManuelStarDo (changes requested):
@rguerra-fccn 
Great improvement, just one issue before we can merge

**[MEDIUM]** `nau_openedx_extensions/xblocks/capa_problem_responses.py:224` (`SKIP_ELEMS = ('description',)`) — the "skip over" allowance when walking backwards from the answer element only accounts for `<description>`, not other structurally-inert siblings that Studio/OLX authors commonly insert for spacing/formatting, most notably a bare `<br/>`.
Risk: For the `optionresponse`/matching case (the whole reason `<div>` recovery exists), if Studio (or a manually-edited/legacy course) ever emits a layout `<br/>` *directly* between an item's `<div>` prompt and its `<optioninput>` — rather than between the previous `<optioninput>` and the next `<div>` (the only ordering the new test `test_recovers_question_from_div_wrapped_prompt` actually exercises) — `getprevious()` will land on `<br/>`, which is neither skipped (`SKIP_ELEMS`) nor a valid `LABEL_ELEMS` match, so recovery silently fails and falls back to the "Question N" placeholder for that specific sub-answer. This is exactly the same class of bug this PR is fixing (a formatting/wrapper element sitting between the prompt and the answer defeating a positional lookup), just one token further back, and it's plausible in hand-edited or older-format OLX even if it wasn't present in the one production course tarball verified in this PR's testing (36 files / 39 answer_ids is a good sample, but not exhaustive across all NAU courses).
Fix: Extend `SKIP_ELEMS` to also skip known non-prompt structural/whitespace elements (at minimum `br`), the same way `description` is already skipped, e.g. `SKIP_ELEMS = ('description', 'br')`. Add a test with a tree shaped like `<div>prompt</div><br/><optioninput id="..."/>` (prompt separated from the answer by a trailing `<br/>`) to lock in the fix and document the assumption.

rguerra-fccn:
Thanks for catching this, @ManuelStarDo! Addressed in `049d535` (pushed to this branch):

**[MEDIUM] `SKIP_ELEMS` too narrow — only `<description>`**

Extended `SKIP_ELEMS` to also skip `<br/>`, `<hr/>`, and XML comment nodes when walking backwards from an `<optioninput>`/answer element looking for its prompt. Comment nodes needed a separate check (`isinstance(candidate.tag, str)`) since lxml represents their `.tag` as a callable rather than a string, so they'd never match a plain name-based `SKIP_ELEMS` tuple.

This is intentionally still a narrow, explicit list of known structural/non-prompt element kinds — not a blanket "skip any element with no text" rule. That broader version would actually be *unsafe*: it would also skip past a genuinely empty `<p>`/`<label>`/`<div>` slot into unrelated shared/"overall instructions" content further back and return the wrong text for that slot.

Added two tests locking in the exact scenario you described (prompt separated from its own answer, not just between sibling sub-answers):
- `test_recovers_question_from_div_wrapped_prompt_with_trailing_br` — `<div>prompt</div><br/><optioninput/>`
- `test_recovers_question_from_div_wrapped_prompt_with_hr_and_comment` — `<div>prompt</div><hr/><!--comment--><optioninput/>`

Re-verified: 22/22 unit tests pass, `pylint`/`pycodestyle`/`isort` clean against CI's exact invocation, and re-ran the 36-file/39-`answer_id` production OLX regression sample — still 39/39 recovered, 0 regressed. CI is green again on this PR.

ManuelStarDo (approved):
Hello @rguerra-fccn 

This PR is good to merge.

Here are a few suggestions:

**[MEDIUM]** `nau_openedx_extensions/xblocks/capa_problem_responses.py:298-303` — Default-label detection is coupled to the exact upstream English string `"Question {}"`
Risk: This assumes `LoncapaProblem.find_question_label`'s internal default template is (and stays) exactly `"Question {}"` translated via `capa_system.i18n.gettext`. If upstream edx-platform ever changes that template's wording, format-placeholder style (e.g. `%s`/`%(num)s` instead of `{}`), or capitalization, `is_default_result` silently becomes always-`False` for non-empty results, and the fix stops recovering the primary reported bug (rich-text nested prompts) — it would still work for the empty/`None` case, but not for the "generic Question N is returned" case, without any test or CI signal catching the drift, since it depends only on live upstream behavior, not on anything in this repo.
Fix: Consider asserting/logging when the reconstructed template doesn't look like what's expected (e.g. warn if `is_default_result` never triggers across N calls in a request), or add a short comment/TODO cross-referencing the exact edx-platform source line this assumes, so a future upstream bump is easier to diagnose if this silently regresses. Not blocking, since the existing fallback behavior is still safe (never worse than before the patch), just worth flagging given the intentionally fragile coupling.

**[LOW / SUGGESTION]** `nau_openedx_extensions/xblocks/capa_problem_responses.py:236` — `SKIP_ELEMS`/`LABEL_ELEMS` are re-created on every call
Risk: None functionally — this is a minor micro-optimization note, not a bug. `find_question_label` is called once per answer in a report loop (up to \~39 answers per course per the PR's own testing), so the cost is negligible.
Fix: If this is ever called at larger scale (e.g. very large courses), hoist `SKIP_ELEMS`/`LABEL_ELEMS` tuples to module-level constants (similar to `_BLANK_JOINED_TEXT_RE` above it) to avoid re-allocating on every call. Purely optional.

**[LOW / SUGGESTION]** `nau_openedx_extensions/xblocks/capa_problem_responses.py:265` — No type hints on new public-ish functions (`_find_label_element_text`, `find_question_label_wrapper`)
Risk: Minor readability/maintainability only.
Fix: Not required — consistent with the rest of the file (`find_correct_answer_text_wrapper`, `extract_choices_wrapper` also have no type hints), so this matches existing project convention. No action needed; noted only for completeness.

**[LOW / SUGGESTION]** `nau_openedx_extensions/tests/test_capa_problem_responses.py` — No test for the interaction between the new `find_question_label` patch and the already-patched `find_correct_answer_text`/`extract_choices` on the *same* problem tree
Risk: Low — each patch is unit-tested in isolation and the PR description states manual end-to-end verification (`generate_report_data()`) was done against real production OLX, which covers this.
Fix: Not blocking given the manual verification described; could optionally add one combined integration-style test asserting all three columns recover correctly for the same synthetic rich-text MCQ tree, for regression safety if one of the three patches is touched independently in the future.

rguerra-fccn:
Thanks for approving, and for the additional suggestions, @ManuelStarDo! Addressed all three actionable ones in `032d0b6` (pushed to this branch):

**[MEDIUM] Default-label detection coupled to the exact upstream `"Question {}"` string**

Added a comment documenting exactly what this assumes (`LoncapaProblem.find_question_label`'s nested `generate_default_question_label()` in edx-platform's `xmodule/capa/capa_problem.py`, currently using `_("Question {}").format(...)`), and what happens if that ever drifts: `is_default_result` would silently become always-`False` for non-empty results, so this patch would stop recovering the "generic Question N" case specifically — it would still recover the empty/`None` case, and is never worse than pre-patch, but the rich-text-prompt fix itself would silently regress with no test/CI signal, since it depends on live upstream behavior.

**[LOW] `SKIP_ELEMS`/`LABEL_ELEMS` re-created on every call**

Hoisted to module-level constants (`_SKIP_ELEMS`, `_LABEL_ELEMS`, `_LABEL_ELEMS_WITH_DIV`), matching the existing `_BLANK_JOINED_TEXT_RE` convention already in the file.

**[LOW] No integration test combining all three patches**

Added `TestCombinedPatchesIntegration`, exercising `extract_choices`, `find_correct_answer_text`, and `find_question_label` together on the same synthetic rich-text MCQ tree, for regression safety if one patch is touched independently later.

The type-hints suggestion was explicitly called out as not needed (matches the file's existing convention) — no change made for that one.

Also did a final sanity pass beyond the unit tests: ran the actual `generate_report_data` code path (the real report generator — `LoncapaProblem` + `find_question_label`/`find_answer_text`/`find_correct_answer_text`) against real production OLX with real submitted answers, for both a rich-text MCQ problem and a `<div>`-wrapped matching/dropdown problem (including a sub-answer separated by `<br/>`). Confirmed the `Question`/`Answer`/`Correct Answer` report columns all resolve to correct real text end-to-end, not just in the isolated unit tests. 23/23 tests pass, `pylint`/`pycodestyle`/`isort` clean, CI green.

PR #159, opened 2026-08-27
feat: add NUTS and CAE4 profile fields, extend employment situation

efortish:
## What this does

Phase 1 (TDC) of [fccn/nau-technical#954](https://github.com/fccn/nau-technical/issues/954): the four extended profile fields, and the logic that keeps a learner out of a course until they have filled them in.

`nif` already existed on `NauUserExtendedModel` but was never added to `NauUserExtendedForm`, so it was invisible at registration. `nuts` (a single merged "NUTS II - NUTS III" field) and `cae4` are new. `employment_situation` keeps its column and gets the new option list.

## The fields

The three controlled lists moved to `custom_registration_form/choices.py`. The database stores short stable codes like `public_senior_technician`, not the display string, so rewording a label is a code change instead of a data migration.

The labels are Portuguese literals and deliberately skip `gettext`. They are exported in `student_profile_info`, which runs inside a Celery task where the active language is not guaranteed, and the reports have to read in Portuguese either way. If English is ever needed there, wrap them and add the catalogs.

**Heads up for review:** the `User` report properties now return `get_FOO_display()`. Site configuration already lists `nau_user_extended_model_employment_situation` in `student_profile_download_fields`, so merging this changes an existing production report: that column goes from `Student` to `Estudante`. Nothing to configure, it just happens.

### The old employment values

They were stored as English display strings and are remapped in place by migration `0015`. "Public service contract" was split upstream into ten "Função Pública" entries, so there is no single value to map it to. It keeps its own code and stays selectable, labelled as the previous option.

That is not tidiness. `partial_update` in `context_extender.py` calls `full_clean()` on the whole model whenever any account field is saved, so dropping that value from `choices` would break account edits for every learner still holding it, on a field they never touched.

## The gate

Three filters, one per way into a course. All three read the same course setting, so a course declares its required fields once.

| Filter | Where it acts |
| --- | --- |
| `FilterEnrollmentRequireProfileFields` | The enrollment. The API answers 403 with the message naming the missing fields. |
| `RequireProfileFieldsOnCourseAbout` | The course about page, replaced by the completion panel. |
| `RequireProfileFieldsOnXBlockRender` | The course content. Staff are let through. |

The third one is the one that matters. The other two are bypassed by anyone already enrolled, either from before the course started requiring the fields or through a bulk enrollment, and that learner would otherwise walk straight into the content.

It uses `RenderCustomResponse` rather than `PreventXBlockBlockRender`, because the latter returns `render_500(request)` and the learner would get a generic error page. Note the xblock view wraps the result in a `Fragment`, so that filter passes markup while the course about one passes an `HttpResponse`. Same exception name, different expectations.

`RequireProfileFieldsOnCourseAbout` does not run when `ENABLE_MKTG_SITE` is on, which is how NAU runs: the about view redirects to the course home before the filter is reached. It is kept because it is the right gate for deployments without a marketing site, but at NAU the content filter does the work.

## Configuration

Which fields block is **per course**, set in Studio under Settings, Advanced Settings, Other Course Settings. This follows the same pattern as the existing `filter_enrollment_require_nif`:

```json
{
    "filter_enrollment_require_profile_fields": ["nif", "nuts", "cae4"]
}
```

A course without the setting does not gate on anything, which is the default. That is deliberate: a course team decides what its own course needs rather than the platform imposing one list everywhere, and it lets the fields be rolled out course by course instead of locking everyone out on the day this ships.

Names resolve against `NauUserExtendedModel` first and then the native `UserProfile`, so `nuts` and `year_of_birth` both work. `nif` goes through `is_nif_valid` rather than being checked for presence, since a stored but invalid NIF should not count as filled. A name that matches neither model is logged and skipped instead of treated as missing: a typo in the advanced settings should not lock a course for everyone.

The filters still need registering in `OPEN_EDX_FILTERS_CONFIG`, and site configuration needs `nuts` and `cae4` in `extended_profile_fields` or the fields exist and are never shown. Both live in [fccn/nau-tutor-configs#318](https://github.com/fccn/nau-tutor-configs/pull/318).

## How to test

Tested against `nau-tutor-configs` with the `nauedu/openedx-dev:nau-teak.master` image, since the account page needs the extension points that only exist in the fork. The gate itself works on either.

1. Mount and install the plugin:
   ```
   bin/tutor mounts add lms:<path>/nau-openedx-extensions:/openedx/nau-openedx-extensions
   bin/tutor dev exec lms pip install -e /openedx/nau-openedx-extensions
   ```
2. `./manage.py lms migrate nau_openedx_extensions`
3. Register the three filters and add `nuts` and `cae4` to `extended_profile_fields`, then restart the LMS. A redis flush is not enough, site configuration is cached in the process.
4. In Studio, put the JSON above in Other Course Settings for a demo course.
5. With a learner who has none of those fields, and who is **not** staff:
   - Try to enroll. The enrollment is refused, and the reason is in the 403 body.
   - Open a unit. The content is replaced by the panel listing what is missing, with a link to the account page.
6. Fill the three fields at `/account/`, reopen the unit, and the content shows.
7. To check the case that matters, enroll the learner first and clear the fields afterwards. They stay out of the content even though they are enrolled.

Switch the browser to Portuguese to see the translated copy. Unit tests: `pytest nau_openedx_extensions/tests/test_pipeline.py nau_openedx_extensions/tests/test_extended_profile_choices.py`.

## What this does not cover

* **The course outline is still visible.** The learning MFE builds it from `/api/course_home/outline/`, and `lms/djangoapps/course_home_api` exposes no `openedx_filters` hook, so there is nowhere to attach one. A blocked learner sees the section titles and cannot open anything. Gating it would need middleware in the plugin or a patch in the fork.
* **The enrollment refusal is invisible in the MFEs.** The API returns the reason as `message` and `localizedMessage`, but neither `frontend-app-learning` nor `frontend-app-learner-dashboard` reads it, so a learner clicking enroll from an MFE sees nothing. That needs a frontend change.
* **The account page does not highlight the missing field.** The panel links to it with the names in a `missing` query parameter, ready for whenever the component reads it.
* **The Educational level translation update** from the issue is not here. It is a native field, so it goes through a translation override, not this model.

## Notes

The panel carries no colours of its own, it uses `btn btn-primary` so the site theme paints it. Copy is translated to pt_PT, including the field labels, which come from the model `verbose_name` (NIF shows as `Nº de Identificação Fiscal`). The Portuguese wording could use a review from someone who speaks it natively.

Two tests pin the language with `translation.override("en")`. They compare English copy, so without it they pass or fail depending on the active locale, which is exactly what happened when the translations were added.


nau_openedx_extensions/templates/profile_completion/_panel.html
```diff
@@ -0,0 +1,39 @@
...
+  <p>${body_text}</p>
+  <ul>
+    % for label in missing_labels:
+      <li>${label}</li>
+    % endfor
+  </ul>
+  <div class="nau-profile-completion-actions">
+    <a class="btn btn-primary" href="${account_url}">${_("Go to my account")}</a>
+    % if course_about_url:
+      <a class="btn btn-secondary" href="${course_about_url}">${_("Back to the course")}</a>
```
dcoa: I think this button is not needed, but if you want to keep it I suggest:


```suggestion
      <a class="btn btn-secondary" href="${course_about_url}">${_("Back to course details")}</a>
```

This message is displayed when I am **in** the course, so if you tell me _Back to the course_ I could understand the modal will just disappear and I will continue my workflow

nau_openedx_extensions/templates/profile_completion/_panel.html
```diff
@@ -0,0 +1,39 @@
...
+</style>
+<section class="nau-profile-completion">
+  <h2>${heading}</h2>
+  <p>${body_text}</p>
+  <ul>
+    % for label in missing_labels:
+      <li>${label}</li>
+    % endfor
+  </ul>
+  <div class="nau-profile-completion-actions">
```
dcoa: Please place the primary button at the end of the modal, this is for better user experience.

For better understanding, you can check https://balsamiq.com/blog/button-design-best-practices/#5-place-primary-actions-where-users-finish-their-task

ManuelStarDo:
This PR reverts:
- https://github.com/fccn/nau-openedx-extensions/pull/146

The same way that https://github.com/fccn/openedx-platform/pull/48 reverts https://github.com/fccn/openedx-platform/pull/45

@efortish Can you please put that info into the PR description?

rguerra-fccn (changes requested):
**Verdict: Request changes.**

Thorough piece of work overall — the `docs/extended_profile_fields.rst` addition is excellent, the migration is properly reversible, and `partial_update()`'s allow-list check closes the mass-assignment gap I'd flagged against the `PROFILE_EXTENSION_FORM` approach in #47/#48. Two things need fixing before this merges, one of them CRITICAL.

### CRITICAL: the course gate silently lets learners through when they have no `NauUserExtendedModel` row at all

`filters/pipeline.py`, `profile_field_is_filled()`:
```python
for source in (getattr(user, "nauuserextendedmodel", None), getattr(user, "profile", None)):
    if source is not None and hasattr(source, field_name):
        return bool(getattr(source, field_name))
return None
```
When a user has no `NauUserExtendedModel` row — anyone who registered before this feature, or via any path that skips `NauUserExtendedForm` (SSO auto-provisioning, admin-created accounts) — `nauuserextendedmodel` is `None`, the loop falls through to `profile`, which doesn't have `nuts`/`cae4`/`employment_situation` either, so the function returns `None`. `missing_profile_fields()` treats `None` as "not a real field, log and ignore" rather than "missing".

Verified live: a fresh user with a `UserProfile` but zero `NauUserExtendedModel` row, course gated on `nuts` only:
```
WARNING ... Course ... requires the profile field 'nuts', which does not exist on NauUserExtendedModel or UserProfile. Ignoring it.
RESULT: enrollment ALLOWED despite nuts being unset
```
`nif` doesn't hit this because it's special-cased through `user.nau_nif`, which already tolerates a missing row. `nuts`/`cae4`/`employment_situation` aren't, and this is exactly the population Phase 1 most needs gated — anyone who hasn't touched these fields yet.

Why the new tests didn't catch it: every test double builds `nauuserextendedmodel=self.FakeModel(**extended_fields)` — always a populated fake, even when a field is explicitly `None`. That models "row exists, field empty" (correctly caught), never "row doesn't exist" (the actual bug). Worth adding a case where `nauuserextendedmodel` is `None`/absent.

**Suggested fix**: check whether `field_name` is a real field on `NauUserExtendedModel` via `_meta.get_field()` before falling through to "doesn't exist"; treat "valid field, no row" as `False` (not filled), not `None`.

### Blocking per fccn/nau-technical#1050

`context_extender.update_account_serializer()` still does a blind `.extend()` — the stale/duplicate `extended_profile` entries issue isn't fixed here yet. @ManuelStarDo already flagged this as required before merging this PR; confirmed it's still present in the current 10-commit state.

### Minor, non-blocking
- `NauUserExtendedForm.Meta.fields` (`forms.py`) lists `"nif"` twice — harmless, but worth a one-line cleanup.
- `dcoa`'s two review comments on `_panel.html` (button copy/placement) are still open in the current diff.

ManuelStarDo (changes requested):
Hello @efortish, good job on the PR. This is solid work.

Requesting changes as I agree with all of @rguerra-fccn and @dcoa 's points.

Also a very minor thing to keep in mind (non-blocking):

**[Note] `nau_openedx_extensions/filters/pipeline.py` — `profile_field_label()` does an in-function import**
```python
def profile_field_label(field_name):
    from django.core.exceptions import FieldDoesNotExist
    from nau_openedx_extensions.custom_registration_form.models import NauUserExtendedModel
```
Not wrong (avoids a potential circular import), but if the CRITICAL fix above ends up needing the same `NauUserExtendedModel._meta.get_field()` lookup in `profile_field_is_filled()`, consider consolidating so the field-existence check lives in one place rather than being duplicated between the label helper and the fill-check helper.

nau_openedx_extensions/templates/profile_completion/_panel.html
efortish: Agreed, it is gone. With the panel rendering over the content there is nowhere useful to go back to, the learner either fills the fields or leaves, so a second button was just noise.

nau_openedx_extensions/templates/profile_completion/_panel.html
efortish: Done, by removing the other button rather than reordering. The primary is now the only action and it sits last in the panel, which is what the article argues for.

efortish:
Both blockers are fixed, plus the minor points.

**The gate letting through learners with no row.** Confirmed, and worse than the report suggests: it was not just `nuts`, it was every field except `nif`, for every learner who has not touched these fields, which is nearly all of them. `profile_field_is_filled()` asked the instance whether the field existed, so no row read the same as no such field, and `missing_profile_fields()` skips those on purpose so a typo in the advanced settings cannot lock a course.

It now asks the model through a `nau_profile_field()` helper that `profile_field_label()` shares, so field existence is decided in one place. @ManuelStarDo, that is your consolidation point, the `_meta.get_field()` lookup is no longer duplicated.

You were also right about why the tests missed it. Every double built a populated fake row, so they covered "row exists, field empty" and never "no row". Added that case.

**fccn/nau-technical#1050.** Fixed here rather than in its own PR: `update_account_serializer()` does not exist on `nau/teak.master`, #146 deleted it, so there is nothing to patch there. It merges by field name now and the model wins, which is the source the gate reads.

One consequence worth flagging: an empty model value overrides a stale `meta` value. That is the point, the page has to show what gates the course, but a learner whose `nif` only ever reached `meta` will see that field go blank on the account page and will have to fill it again. It was never going to let them into a gated course anyway.

**Minor.** Duplicate `"nif"` in `Meta.fields` gone. @dcoa's two comments answered and resolved, the secondary button is already removed so the primary is the only action and sits last. PR description now says what this reverts.

103 tests passing, lint clean.

rguerra-fccn (approved):
**Verdict: Approve.**

Re-reviewed after the latest 3 commits, and re-verified both of my previous CRITICAL findings live (not just read the diff) — both are genuinely fixed.

**No-row gating bypass — fixed and verified.** `profile_field_is_filled()` now asks the model itself (via the new shared `nau_profile_field()` helper, also used by `profile_field_label()` — nice consolidation, no more duplicated `_meta.get_field()` lookups) whether a field exists, instead of asking the user's instance. A missing `NauUserExtendedModel` row now correctly reads as "not filled" rather than "not a field, skip it." Re-ran my exact original reproduction live: a user with zero rows, course gated on `nuts` → enrollment is now correctly blocked (previously silently allowed). The new test explicitly constructs `nauuserextendedmodel=None`, which is exactly the case none of the old tests modeled — good root-cause fix to the test gap too.

**fccn/nau-technical#1050 (stale duplicate `extended_profile` entries) — fixed and verified.** `update_account_serializer()` now merges by `field_name` into a dict, with the model overwriting meta, instead of blind `.extend()`. Re-ran my exact original reproduction: no duplicate entries, model value wins for every overlapping field (`nif`, `nuts`, `employment_situation`, `cae4`) — matches "the gate reads the model" exactly. Good call fixing this here directly rather than waiting on the tracking issue's own PR, since `update_account_serializer` only exists on this branch/repo.

**Minor items confirmed resolved**: duplicate `"nif"` in `NauUserExtendedForm.Meta.fields` is gone; @dcoa's two UX comments are addressed (verified the template directly — one primary button, correctly positioned last).

**Full end-to-end regression check**, combining this PR with the already-merged openedx-platform#48: edited `nif` via `update_account_settings` → correctly creates/updates the `NauUserExtendedModel` row through the restored `NAU_STUDENT_ACCOUNT_PARTIAL_UPDATE` bridge → the course gate correctly sees `nif` as filled while `nuts` (never touched) correctly remains flagged as missing. The whole Phase 1 stack (this PR + #48) is coherent end-to-end.

No new issues found. Good to merge.

ManuelStarDo (approved):
Great job @efortish LGTM, just need to rebase, clean and ready to merge 🚀

Can you please:
1. Squash the commit history into ~2 commits: one for restoring the account extension points, one for the NUTS/CAE4 fields + gate filters (feel free to use your judgment if you see it differently, just need a clean git history)
2. Rebase onto nau/teak.master 
4. Add Related to: fccn/nau-technical#954 at the end of each final commit message and the PR description

efortish:
Done @ManuelStarDo, thanks for the review.

Squashed to two commits, split the way you suggested:

- `34df2fe` the NUTS/CAE4 fields, the migration, the choice lists and the three gate filters
- `2803416` the account extension points restore, including the allowlist check on the write side and the #1050 serializer merge

Both carry `Related to: fccn/nau-technical#954` and it is at the end of the PR description too. The second one also closes fccn/nau-technical#1050.

On the rebase: the branch was already level with `nau/teak.master`, zero commits behind, so it brought nothing new. What it did change is that the old merge commit is gone and the history is linear now.

efortish:
Added one more commit, `0cb4df6`, docs only. No code change, so the approvals still cover what runs.

I re-read `docs/extended_profile_fields.rst` against the code and found three gaps:

The gate section still described the old instance-based lookup, the shape that had the bug @rguerra-fccn found. It never said that field existence is decided on the model, so a learner with no `NauUserExtendedModel` row counts as not having filled it. That is the whole point of the fix and it was not written down anywhere.

Nothing said `extended_profile` is filled from two sources and merged with the model winning. The consequence belongs in writing for whoever picks up a support ticket: a learner whose value only ever reached `UserProfile.meta` finds the field blank on the account page and has to enter it again.

`NAU_ACCOUNTS_CC_VISIBLE_FIELDS` and `extended_profile_fields` were documented separately with no mention that they have to agree. A field in only one renders and then silently fails to save.

It does make this three commits instead of two, @ManuelStarDo. I kept it separate rather than amending so you can see exactly what changed without me rewriting a branch you already approved. Happy to fold it into the second commit if you would rather have two, just say so.

efortish:
Added `84db3ad`, a setup guide at `docs/profile_fields_setup.rst`. Docs only again.

The existing page explains what the fields are and why the design is what it is, but it never says what to actually do, and the per-course part, which is the bit most people need, was three paragraphs buried in the middle of a reference page.

The new one is the how-to: the three pieces that have to be in place and what breaks without each, how to check the image actually carries the #48 hooks, the migration, the settings, the filter registration, and a shell snippet to confirm the account API is returning the fields.

The per-course section has the eight valid field names and which model each lives on, where exactly it goes in Studio, what the learner hits at each of the three entry points, and why invites are refused for a gated course.

It closes with a troubleshooting table built from the ways we actually got this wrong while building it. The one that will cost someone an afternoon: a field listed in `extended_profile_fields` but not in `NAU_ACCOUNTS_CC_VISIBLE_FIELDS` renders on the account page and then silently refuses to save.

@ManuelStarDo this makes four commits now, two code and two docs. Same offer as before, happy to squash the two docs ones together or fold them into the second commit, whatever keeps you happy. I would rather ask than rewrite the branch again.

rguerra-fccn:
Hello @efortish, this PR has already been merged, so we cannot see these new commits. Can you create a new PR with these commits?

efortish:
@rguerra-fccn 

I did not noticed that my apologies, I'll open a new PR with the documentation :+1:

